Issue #17, opened 2020-03-13
[CKEditor 5] Upcast converter tries to convert all span view elements


markcharyk:
If you try to use the `mathtype-ckeditor5` plugin in an editor with `span`s that do not have `class="ck-math-widget`, the plugin will try to upcast them anyways.

So to repro, create a simple custom plugin with the following conversion:
```
this.editor.conversion.for('upcast').elementToElement({
  view: {
    name: 'span',
    attributes: {
      'data-role': 'not-math'
    }
  },
  model(viewElement, modelWriter) {
    return modelWriter.createElement('span');
  }
});
```
If you install both plugins and try to call `editor.setData('<p>Here is an <span>error</span>.</p>);` you end up with an error of:
```
 viewElement.getChild(...).getAttribute is not a function
```

xaviripo:
This was fixed in ae343f8.

PR #18, opened 2020-03-13
Fix a typo in the ckeditor5 plugin upcaster

markcharyk:
This will limit upcast conversions only to elements with class="ck-math-widget"

This fixes https://github.com/wiris/html-integrations/issues/17


xaviripo:
This was fixed in ae343f85636d011546a871e5306cf1866b9f68db.

PR #28, opened 2020-04-02
Plugins 1329 fix

xaviripo:
Branch PLUGINS-1329 with cleaner history.

### DIFF
--- a/packages/mathtype-ckeditor5/src/integration.js
+++ b/packages/mathtype-ckeditor5/src/integration.js
@@ -254,6 +254,10 @@ export default class CKEditor5Integration extends IntegrationModel {
             );
         }
 
+        /* Due to PLUGINS-1329, we add the onChange event to the CK4 insertFormula.
+        We probably should add it here as well, but we should look further into how */
+        // this.editorObject.fire('change');
+
         return returnObject;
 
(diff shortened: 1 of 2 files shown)


packages/mathtype-ckeditor5/src/integration.js
```diff
@@ -254,6 +254,9 @@ export default class CKEditor5Integration extends IntegrationModel {
             );
         }
 
+        // Fire the 'change' event.
+        // this.editorObject.fire('change');
+
```
xaviripo, line 259: Do we leave this?

dcanetma: We added this callback to the version 4 of the plugin.
It may be a good a idea to do the same for the next version 
of the plugin. Otherwise, users who would upgrade and are
using this feature would miss it. 

But we decided to comment it so it'd need to be properly tested.

xaviripo: The integration for CK5 is vastly different. I believe we should leave this change out, or at least add a comment explaining why it's there, so as to avoid any confusion when we see this piece of code in the future and might think that it used to be uncommented.

PR #35, opened 2020-05-11
Plugins 1340

carlawiris:


CONTRIBUTING.md
```diff
@@ -0,0 +1,25 @@
...
+
+1. **Code contributing**. First, clone this repository, create a branch from master, code and submit it for a PR. Then, we'll review, discuss, and merge changes as needed. You can contribute by:
+
+    * Adding a missing or new feature.
+    * Solving an existing bug.
+    * Submitting any code you think that may improve this project.
+
+2. **Reporting issues**. You can report an error by:
+    * Sending an email to team.support.europe@wiris.com.
+    * Opening a GitHub issue on this repository.
```
xaviripo, line 13: We should be more explicit about this being two mutually exclusive options. Maybe add a ", or" at the end of the first option.

PR #36, opened 2020-05-11
PLUGINS-1338

carlawiris:
I would remove the reporting part, for now, since it is not up to us, and add it in the future after we have made a decision about this section.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -33,13 +33,3 @@ Harassment is not tolerated and must be reported. Harassment includes:
 Every individual asked to stop any kind of harassment is expected to comply immediately.
 
 We encourage to report any harassment suffered.
-
-## Reporting 
-
-If you are the subject of harassment, notice that someone else is being harassed, or have any other concerns, please contact us by email ____________@wiris.com as soon as possible. All reports will be handled with discretion. In your report please include:
-
-* Your contact information.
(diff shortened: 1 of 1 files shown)


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,45 @@
...
+
+- Virtual harassment.
+
+Every individual asked to stop any kind of harassment is expected to comply immediately.
+
+We encourage to report any harassment suffered.
+
+## Reporting 
+
+If you are the subject of harassment, notice that someone else is being harassed, or have any other concerns, please contact us by email ____________@wiris.com as soon as possible. All reports will be handled with discretion. In your report please include:
```
xaviripo, line 39: We are still missing a mail. We should talk with someone from HR.

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,45 @@
...
+
+* Your contact information.
+* Names (real, nicknames, or pseudonyms) of any individuals involved. If there are additional witnesses, please include them as well. Your account of what occurred, and if you believe the incident is ongoing. If there is a publicly available record (e.g. a mailing list archive or a public IRC logger), please include a link.
+* Any additional information that may be helpful.
+
+After filing a report, a representative will contact you personally. If the person who is harassing you is part of the response team, they will recuse themselves from handling your incident. A representative will then review the incident, follow up with any additional questions, and make a decision as to how to respond. We will respect confidentiality requests for the purpose of protecting victims of abuse.
```
xaviripo, line 45: As much as I'd like this to work, we should talk with HR to ensure that we can guarantee this. With these statements we are committing ourselves, we should make sure as a team/company we have the time, resources and formation to deal with this. If you want you can bring this up in the daily meeting so we can plan how to deal with it.

PR #45, opened 2020-06-08
React+froala3

carlawiris:


carlawiris:
> The demo looks great and works fine. Congrats!!!
> 
> Before accepting changes, please have a look to this next list of suggested changes.
> 
> **Suggested changes:**
> 
> 1. When running the demo this next warning message appears: "Warning: componentWillMount has been renamed, and is not recommended for use. See https://fb.me/react-unsafe-component-lifecycles for details."
>    
>    * Is it possible to migrate quickly from this Legacy Lifecycles instructions to the new aliases available? If so, it may be a good idea to do it.
> 2. On the README file:
>    
>    * Add the  "$ npm install" command to the 'How to run the demo' section; (like the angular react does).
>    * I'm missing some instructional comments on the code of the 'How to add MathType to TinyMCE from scratch' section. I suggest you to have a look a the angular react README.
> 3. On the package.json file:
>    
>    * Upgrade jquery package to `"^3.5.1"` to fix a high risk vulnerability
>    * Update 'author' in the package.json file ;)
>    * Set 'description' value in the package.json file: _"A simple ReactJS App integrating WIRIS MathType..."_

The point number one is a warning comming from Froala side, we do not cause it more than when using froala packages. We assume they'll solve this in a future and keep working as usual.

PR #53, opened 2020-07-01
Plugins 1335

carlawiris:


demos/html5/ckeditor5/README.md
```diff
@@ -0,0 +1,19 @@
...
+$ npm start
+$ npm deploy
+$ npm build-dev
+```
+
+## Development mode
+
+To run the demo in the development mode, some changes have to be done:
+
+1. Replace the content of the "@wiris/mathtype-ckeditor5" to **"file:../../../packages/mathtype-ckeditor5/wiris-mathtype-ckeditor5-7.20.0.tgz"** but with the correct version. Now it is 7.20.0 but it can change in the future. Check the current version in the plugin package.json.
```
xaviripo, line 18: Here I'd put

>**"file:../../../packages/mathtype-ckeditor5/wiris-mathtype-ckeditor5-x.y.z.tgz"** where x.y.z represents the current version, e.g. wiris-mathtype-ckeditor5-7.20.0.tgz. Check the current version in the plugin package.json.

demos/html5/ckeditor5/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+Here are some instructions to work with this demo and with lerna, as it is complicated to do the same way the other demos do. That is because CKEditor 5 does not let to have duplicate modules, and this has interferences whit the demo in the development mode. 
+
+## Production mode
+
+Use one of the following commands to run the demo in the production mode:
+```sh
+$ npm start
+$ npm deploy
+$ npm build-dev
```
xaviripo, line 11: All three are interchangeable? In that case, I would just write one instead of all three.

lerna.json
```diff
@@ -1,6 +1,16 @@
...
+    "packages/*",
+    "demos/html5/ckeditor4",
+    "demos/html5/froala2",
+    "demos/html5/froala3",
+    "demos/html5/generic",
+    "demos/html5/tinymce4",
+    "demos/html5/tinymce5",
+    "demos/html5/ckeditor5",
+    "demos/angular/*",
+    "demos/react/*"
```
xaviripo, line 12: Is this change on purpose? Will this branch solve the issues we had with the Jenkins machines when running the tests?

carlawiris: this branch solves the issue but these are not necessary, my mistake

demos/html5/ckeditor5/README.md
carlawiris: Actually, i've found a way to improve the package.json scripts so the readme won't be necessary

PR #55, opened 2020-07-07
Fix formula generator in parser

ilyasgaraev:
The `forEach()` method executes a provided function once for each array element. It means that `const param = renderParams[key];` line is redundant, because `key` returns parameter name.

This bug causes the issues with formula – `dataMd5` is always empty and `dataObject.formula` is always equal to "d41d8cd98f00b204e9800998ecf8427e".


carlawiris:
Hi @ilyasgaraev, thank you for the report. 

The truth is that we have worked on this through another client's report, in the issues section, which proposes to solve exactly the same problem. We have already created a pull request for this and it is accepted by the whole team. So the fix will be available in the next version.

carlawiris:
We have released a new version that includes this fix, it is the 7.26.0

PR #85, opened 2020-08-28
KB-4027: Mathtype crash when using language as an object

carlawiris:
This code solves the crash when the language definition is an object. 
The code has changed tot take into account the the language as a possible object.
It has to be tested that the c ode works with the old language definition and with the language definition as an object (more details in the card 4027).
The tests just takes into account the current language of the demo. More tests will be added when the ckeditor5 stops crashing with the latest version


packages/mathtype-ckeditor5/src/integration.js
```diff
@@ -37,8 +37,16 @@ export default class CKEditor5Integration extends IntegrationModel {
...
+        // Returns the CKEDitor instance language taking into account that the language can be an object.
+        const languageObject = this.editorObject.config.get( 'language' );
+        if (typeof (languageObject) === 'object') {
+            // eslint-disable-next-line no-prototype-builtins
+            if (languageObject.hasOwnProperty('ui')) {
+              return languageObject.ui;
+            } else return 'en';
+          } 
+        return languageObject;
+        // return this.editorObject.config.get( 'language' );
```
xaviripo, line 49: Maybe remove this commented line of code.

xaviripo (changes requested):

dcanetma (commented):
Good work, Carla!

I'm only having a question about the demos. I saw you fixed the 'html5/ckeditor5' demo.

And I was wondering, if we shouldn't review the 'angular/ckeditor' and 'react/ckeditor' versions and updated them to the "^7.22.0" version of "@wiris/mathtype-ckeditor5", also.

What do you think? 

Thanks!

dcanetma:
Nevermind the comment on my review. 

Now I see this other pull request, that updates the dependency to the last version of "@wiris/mathtype-ckeditor5". :+1: 

> https://github.com/wiris/html-integrations/pull/86/

packages/mathtype-ckeditor5/src/integration.js
manuelwiris: +1

manuelwiris (approved):

PR #117, opened 2020-09-21
[CKEditor 5] Removing MathType equation which is being edited results in a crash

carlawiris:
## How to reproduce
See original issue: https://github.com/wiris/html-integrations/issues/94

## Description
The code for this pull request solves a problem with the wiris modal in the ckeditor5: appears an error when deleting a formula on the editor that is being edited in the wiris modal.

In order to accept this review you must check that the code is reliable, that it follows good practices, that it solves the problem mentioned, and that it does not break other aspects of the plugin.

### DIFF
--- a/packages/mathtype-ckeditor5/src/integration.js
+++ b/packages/mathtype-ckeditor5/src/integration.js
@@ -254,11 +254,18 @@ export default class CKEditor5Integration extends IntegrationModel {
             } );
 
         } else {
-            returnObject.node = this.editorObject.editing.view.domConverter.viewToDom(
-                this.editorObject.editing.mapper.toViewElement(
-                    this.insertMathml( mathml )
-                ), windowTarget.document
-            );
+            try {
+                returnObject.node = this.editorObject.editing.view.domConverter.viewToDom(
+                    this.editorObject.editing.mapper.toViewElement(
+                        this.insertMathml( mathml )
+                    ), windowTarget.document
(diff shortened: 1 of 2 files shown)


dcanetma (changes requested):
Try to provide more context for the PR, please.

packages/mathtype-ckeditor5/src/integration.js
```diff
@@ -254,11 +254,18 @@ export default class CKEditor5Integration extends IntegrationModel {
...
-            );
+            try {
+                returnObject.node = this.editorObject.editing.view.domConverter.viewToDom(
+                    this.editorObject.editing.mapper.toViewElement(
+                        this.insertMathml( mathml )
+                    ), windowTarget.document
+                );
+            }  catch (e) {
+                const x = e.toString();
+                if((x).includes("CKEditorError: Cannot read property 'parent' of undefined")) {
```
xaviripo, line 265: We are trying to follow CKEditor 5 conventions for this plugin. In particular, try to leave a single whitespace after `(` and before `)`, e.g. `function( argument )`. Same for `[`, `]`, `{`, and `}`. See: https://ckeditor.com/docs/ckeditor5/latest/framework/guides/contributing/code-style.html

We are not too strict with this, though, but it's good to have in mind for consistency.

xaviripo (approved):

Issue #112, opened 2020-09-13
modelWriter.createElement() is not a function


Arthur944:
I got this error every time I tried to use the plugin. Solution was to change plugin.js line 219:
```
- return modelWriter.createElement('mathml', {...
+ return modelWriter.writer.createElement('mathml', {...
```

xaviripo:
Hi!

We are going to take a look at this issue as soon as possible. Could you please offer a little bit more detail? For example, specific versions of the CKEditor packages and the MathType plugin, and what steps are necessary to reproduce the error?

Thank you.

carlawiris:
We have released a new version that solves this bug, it is the 7.26.0

PR #123, opened 2020-10-13
Fix CKEditor 5 "modelWriter.createElement() is not a function" error

xaviripo:
## Description

This pull request fixes issue #112, which throws an error when writing after a MathType formula in CKEditor 5.

## How to reproduce

1. Open a CKEditor 5 HTML5 demo
2. Write some text, followed by a space
3. Add a MathType equation, followed by a space
4. Type some text

Expected: the text is inserted, no error is thrown.
Actual: the text appears to be inserted, but an error is thrown, and the CKEditor engine doesn't actually process the inserted text.

For more details see #112.

### DIFF
--- a/demos/html5/ckeditor5/package-lock.json
+++ b/demos/html5/ckeditor5/package-lock.json
@@ -1568,6 +1568,15 @@
             "p-locate": "^4.1.0"
           }
         },
+        "p-limit": {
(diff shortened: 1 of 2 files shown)


dcanetma (approved):
Ready to merge

carlawiris (approved):

PR #126, opened 2020-10-20
KB-2971: Fix CKEditor5 not rendering LaTeX when updating $$latex$$ to an image

xaviripo:
## Description

This pull request fixes an issue with the CKEditor 5 plugin.

When calling `editor.getData()` (or when clicking the "Update" button in the HTML5 demos), any formulas written out as LaTeX expressions (e.g. "$$x^2$$") are not rendered as MathML.

## How to reproduce

1. Open a CKEditor 5 HTML5 demo
2. Write some text with a LaTeX expression, e.g. "$$x^2$$"
3. Click "Update"

Expected: the LaTeX is rendered as a formula image. 
Actual: the LaTeX stays as is.


packages/mathtype-ckeditor5/src/plugin.js
```diff
@@ -21,6 +21,7 @@ import Listeners from '@wiris/mathtype-html-integration-devkit/src/listeners';
 import IntegrationModel from '@wiris/mathtype-html-integration-devkit/src/integrationmodel.js';
 import MathML from '@wiris/mathtype-html-integration-devkit/src/mathml.js';
 import Latex from '@wiris/mathtype-html-integration-devkit/src/latex';
+import Constants from '@wiris/mathtype-html-integration-devkit/src/constants.js';
```
dcanetma, line 24: Maybe this import is not necessary.

dcanetma (approved):
Ready to merge.

I would only add that I wasn't able to test the angular and react demos with the local package. 
Imho, it's out of the scope of this pull request and context, though.

carlawiris (approved):

PR #127, opened 2020-10-27
Set Mathtype to accept language property though MathType parameters

carlawiris:
Language property was not accepted through MathType parameters in mathtype plugins.
This pull request changes the definition of the language on the mathtype with the following priorities:

1. Sets the value of the language if this property is defined in MathtypeParameters.
2. Sets the value of the language if the editor language is defined.
3. If none of the above, gets the browser language.

The points mentioned above are those that should be checked if they work as expected. The demos are prepared to reproduce the behavior and the steps are the following:

1. Choose a demo.
2. In the app.js file, there is code to define the language both behind MathtypeParameters and the way the editor defines a language.
3. Comment out and uncomment the mentioned lines to test that the language works according to the priorities that have been indicated at the begining.

There is a special editor, which is the generic one. In this editor, part of the initialization code of the mathtype has been modified so that it accepts the mathtype parameters, so the code is quite different from the other editors but works with the same standards.

This pull request solves a support issue.

### DIFF
--- a/demos/html5/ckeditor4/package-lock.json
+++ b/demos/html5/ckeditor4/package-lock.json
@@ -5674,11 +5674,6 @@
         "@xtuc/long": "4.2.2"
       }
     },
-    "@wiris/mathtype-ckeditor4": {
-      "version": "7.24.0",
-      "resolved": "https://registry.npmjs.org/@wiris/mathtype-ckeditor4/-/mathtype-ckeditor4-7.24.0.tgz",
-      "integrity": "sha512-upFaKjhsDZ/OcWbMjsWKvDLcU+galagY12Zt//VHF+XZCv3KHT0wzvTPaHJnzMTEtXAEh/uUL1aOJVaATQQB6A=="
-    },
     "@xtuc/ieee754": {
       "version": "1.2.0",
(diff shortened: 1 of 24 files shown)


xaviripo:
I uncommented the following lines in the CKEditor 4 demo:

```js
  language: 'de',
  mathTypeParameters: {
    editorParameters: { language: 'es' }, // MathType config, including language
  },
```

Then, I've opened the demo with the usual steps:

```sh
npm install && npm start && cd demos/html5/ckeditor4 && npm install && npm run build-dev
```

 Both CKEditor and MathType appear in German to me.

dcanetma (approved):
Ready to merge.

> _The implementation works fine and covers all cases for all HTML editors.
Good work!_

I've found two issues that I've decided to fix directly since they were quite straight-forward.

1. Fix language setting on tinymce demo: Localization values on tinymce use the form "fr_FR"
to set a different language to the editor's UI.

> _The demo was using the 'fr' format and therefore it was not being recognized by TinyMCE._

2. Normalize deprecated feature message: I've used the same format of deprecated message when 'wirisformulaeditorlang' is informed by the user. 

> _I think it's good idea to be consistent on the console warn messages we send to our users._

xaviripo (approved):

dcanetma (approved):
Ready for Q/A

PR #131, opened 2020-11-12
Create tests to check the language expected definition

carla-at-wiris:
This task includes tests to check the correct resolution of the getLanguage() function in all the mathtype plugins.

There are three main things to be checked:
1. The code of the tests, its settings and their own documentation.
2. The correct execution of the tests.
3. The changes applied on the mathype plugins code that treats the language before calling the constructor.

The configuration files of the tests can be found in the mathtype plguins folder (froala, generic, ...)  and, inside the test folder, there are all the tests created.

To execute them, you can follow the instructions on the README file in the root of the project.
They can be executed one by one inside each mathtype plugin folder, or all together from the root of the project.

### DIFF
--- a/packages/mathtype-ckeditor4/test/unit-language-editorLanguageDefined.test.js
+++ b/packages/mathtype-ckeditor4/test/unit-language-editorLanguageDefined.test.js
@@ -1,16 +1,19 @@
+// Import the CKEDITOR 4 instance to use it un the construction of the instance
 import * as CKEDITOR from '../node_modules/ckeditor4/ckeditor';
 import { CKEditor4Integration } from '../plugin.src'
 
+// Define a global integration to be assigned
 let integration;
 
 /**
- * This tests is focused on check the mathtype language definition through the mathtypeParameters
+ * This test is focused on checking the mathtype getLanguage function through the
+ * editor language definition
  */
-describe('Test the language property. TAG = Language',
+describe('CKEditor4: Test the language property. TAG = Language',
   () => {
     /**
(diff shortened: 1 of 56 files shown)


packages/mathtype-ckeditor4/test/unit-language-editorLanguageDefined.test.js
```diff
@@ -0,0 +1,46 @@
...
+import { CKEditor4Integration } from '../plugin.src'
+
+// Define a global integration to be assigned
+let integration;
+
+/**
+ * This test is focused on checking the mathtype getLanguage function through the
+ * editor language definition
+ */
+describe('CKEditor4: Test the language property. TAG = Language',
```
xaviripo, line 12: Maybe we could use different descriptions for each of the tests, e.g.
"Test the language property when passing only the browser language"
"Test the language property when passing different browser and MathType parameters language"
etc.

xaviripo (approved):

dcanetma (approved):
👏 Good job!
Ready to merge

packages/mathtype-ckeditor4/test/unit-language-editorLanguageDefined.test.js
carlawiris: This type of description is in the test itself and the general description I have put is this so that they are not the same

xaviripo: Okay, I took a second look at the test reports and indeed all the info is there. Great!

Issue #19, opened 2020-03-15
[Parser] Copy for dataMd5 in Parser.createShowImageSrc is wrong


glasses618:
In function `Parser.createShowImageSrc`, the copy for `dataMd5` is wrong,
https://github.com/wiris/html-integrations/blob/139dc19edaa6b1db865ce9e5096369445277f73d/packages/mathtype-html-integration-devkit/src/parser.js#L320-L328

it should be:

```diff
--- a/packages/mathtype-html-integration-devkit/src/parser.js
+++ b/packages/mathtype-html-integration-devkit/src/parser.js
@@ -320,8 +320,7 @@ export default class Parser {
   static createShowImageSrc(data, language) {
     const dataMd5 = [];
     const renderParams = ['mml', 'color', 'centerbaseline', 'zoom', 'dpi', 'fontSize', 'fontFamily', 'defaultStretchy', 'backgroundColor', 'format'];
-    renderParams.forEach((key) => {
-      const param = renderParams[key];
+    renderParams.forEach((param) => {
       if (typeof data[param] !== 'undefined') {
         dataMd5[param] = data[param];
       }
```
because `renderParams` is an array.

This bug causes every call on showimage API is always with the same and wrong parameter of formula because `dataMd5` is always `[]`.
https://github.com/wiris/html-integrations/blob/139dc19edaa6b1db865ce9e5096369445277f73d/packages/mathtype-html-integration-devkit/src/parser.js#L339-L343

carlawiris:
Hi, thank you for your report.

This will be solved in our next release

carlawiris:
We have released a new version that solves this bug, it is the 7.26.0

glasses618:
Good job!

PR #132, opened 2020-11-13
Copy for dataMd5 in Parser.createShowImageSrc is wrong

carlawiris:
The problem is that the dataMd5 always returns [] so the function does not assign the properties to the formula.

It has to be checked that the issue is solved and the formulas are still correctly rendered and have assigned the properties correctly. It can be checked by debugging the code running a Froala demo, for example, in a development mode.

This pull request solves the github issue https://github.com/wiris/html-integrations/issues/19.

### DIFF
--- a/packages/mathtype-html-integration-devkit/src/parser.js
+++ b/packages/mathtype-html-integration-devkit/src/parser.js
@@ -320,8 +320,7 @@ export default class Parser {
   static createShowImageSrc(data, language) {
     const dataMd5 = [];
     const renderParams = ['mml', 'color', 'centerbaseline', 'zoom', 'dpi', 'fontSize', 'fontFamily', 'defaultStretchy', 'backgroundColor', 'format'];
-    renderParams.forEach((key) => {
-      const param = renderParams[key];
+    renderParams.forEach((param) => {
       if (typeof data[param] !== 'undefined') {
         dataMd5[param] = data[param];
       }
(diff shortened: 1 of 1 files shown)


packages/mathtype-html-integration-devkit/src/parser.js
```diff
@@ -320,8 +320,7 @@ export default class Parser {
   static createShowImageSrc(data, language) {
     const dataMd5 = [];
```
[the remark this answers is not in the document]
carlawiris: You're right. The change has been updated

xaviripo (approved):

packages/mathtype-html-integration-devkit/src/parser.js
```diff
@@ -318,10 +318,9 @@ export default class Parser {
    * @returns {Object} JSON object containing showimage response.
    */
   static createShowImageSrc(data, language) {
-    const dataMd5 = [];
+    const dataMd5 = {};
```
dcanetma, line 321: I'm probably wrong, but this variable shouldn't be defined using 'let' instead of 'const'?

dcanetma (approved):
Ready to merge.

ilyasgaraev:
Hello @carlawiris! I opened PR with the same fix and still waiting for merge 😉 https://github.com/wiris/html-integrations/pull/55

packages/mathtype-html-integration-devkit/src/parser.js
carlawiris: True

PR #135, opened 2020-11-27
Define and implement an effective node.js dependency management for the MathType web plugins

carlawiris:
This pull request solves improves the way the dependencies are treated in the demos and the packages.
There are also two documents that explain what are the dependencies and the guides to treat them.

You will have to check that every dependency in every demo and package is in the correct place. It can also be checked that the demo works, but it's not mandatory.
The readme files about dependencies have to be approved too, checking that they tell the truth and that there are no rules missing.


dcanetma (changes requested):
Too much to discuss for a normal peer review.

PR #157, opened 2021-02-23
Solve CK5+windows+build-dev issue

carlawiris:
Solved by adding a script in the package.json CK5 file
that can be executed in all the SO but it's slower than
the regular one.

Issue reproducable just in the CKEditor5 + html5 demo with the npm run build-dev command.
Now, on windows, instead of using npm run build-dev, it has to be used npm run build-dev-windows.
It has to be tested that this command works as expected and that the script created for that is correct.

Issue found by QA team.


xaviripo (approved):

dcanetma (changes requested):
Review if this procedure it's necessary for react and angular demos on ckeditor5.

Validate the code works fine after linting.

dcanetma (approved):

PR #176, opened 2021-03-10
Changing from MT to CT while editing a formula causes a loss of information

carlawiris:
This pull request fixes KB-10222, which states that there's a loss of information when changing from MT to CT and from CT to MT when editing an existing formula in Froala2 and 3, and Generic editors.

### Context
A user opens an existing formula to start editing it and changes the editor from MT to CT, this action generates the loss of the formula the user was working on in the editor

### Why this happens
This is because the froala editor and the generic editor does not keep the existing formula to edit selected while we are in the MT or CT editor.

### Proposed solution
The proposed solution, included in this PR, consist in assign to a temporal register on the current instance, the image of the existing formula that is going to be opened to be edited. Then, every time a new formula is going to be opened, it has to be checked if exists or not the temporal image, in the case that it exists, it's going to be called the openExistingFormula method, if not, the openNewFormula method will be launched. The register is cleared every time the cancer or the insert button are clicked.

### DIFF
--- a/packages/mathtype-froala/wiris.src.js
+++ b/packages/mathtype-froala/wiris.src.js
@@ -149,6 +149,10 @@ export class FroalaIntegration extends IntegrationModel {
      * @param {HTMLElement} element - DOM object target.
      */
     doubleClickHandler(element) {
+        // Save a image to a temporal register to detect when we want to
+        // change between MT and CT.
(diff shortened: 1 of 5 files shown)


dcanetma (approved):
👍 
Looks super good!
Ready to merge.

xaviripo (approved):

PR #179, opened 2021-03-10
Add carret to html-integration-devkit plugins dependency

carlawiris:
This pull request fixes KB-10335, which states that there's a missing carret in the html-integration-devkit dependency on the mathtype plugins to let lerna work the expected way.

### Context
In all the packages of the plugins of mathtype, there's a dependency, html-integration-devkit that sould have a caret (^) in the beggining. This missing caret makes the lerna arquitecture not working when trying to link the dependency to the local package.

### Proposed solution
Add the missing carret on the html-integration-devkit dependency on all the mathtype plugins (except tinymce4 and 5, it is in another PR)

### DIFF
--- a/packages/mathtype-ckeditor4/package.json
+++ b/packages/mathtype-ckeditor4/package.json
@@ -34,7 +34,7 @@
   "devDependencies": {
     "@babel/core": "^7.0.0",
     "@babel/preset-env": "^7.0.0",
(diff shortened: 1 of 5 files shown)


xaviripo (approved):

dcanetma (approved):
👍

PR #195, opened 2021-03-23
KB-10749: Fix html5/tinyMCE* to use local version npm

dcanetma:
- Fix html5/tinyMCE4 to use local version npm.
- Fix html5/tinyMCE4 to use local version npm.
- Bump html5/tinymce5 demo version to 0.0.1.
- Bump html5/tinymce4 demo version to 0.0.1.
- Remove uglify of demo code to limit vulnerabilities.

#taskid 10749


xaviripo (approved):

demos/html5/tinymce5/index.html
```diff
@@ -4,7 +4,7 @@
     <head>
         <title> Demo Tinymce5 </title>
         <meta http-equiv="content-type" content="text/html; charset=UTF-8">
-        <script src="https://cdn.tiny.cloud/1/no-api-key/tinymce/5/tinymce.min.js" referrerpolicy="origin"></script>
```
carlawiris, line 7: Do we know if it's better to add the tinymce dependency from the node modules than the original tinymce cloud?

dcanetma: I believe so.

Imho it's much appropiate to use the tinymce package as defined on the project package.json file:

For example, this is the current dependency:

```json
  "dependencies": {
    "@wiris/mathtype-tinymce4": "^7.0.0",
    "tinymce": "^4.0.0"
  },
```

PR #237, opened 2021-05-04
chore(package.json): send only 'plugin.min.js' to npm

rivajunior:
Without specifying which files should be sent to npm, all files were being sent.  But the only needed file is `plugin.min.js`, since the `global.js` is referencing a package that is in devDependencies, not in dependencies.


xaviripo:
Hi, thanks for the proposal!

You are right that when using the minified code, only `plugin.min.js` is needed. Nonetheless, we offer the unminified source code together with the Webpack configuration needed to minify it, for users that want to minify and transpile the library themselves as part of their own application pipeline. Thus, all the included files are required.

PR #245, opened 2021-05-10
Fix and validate the Node.js project configuration

xaviripo:
There have been issues and inconsistencies in the different Node.js projects inside the html-integrations repository (root, package, and demos).

Some of the identified issues are:

- npm scripts that do not run or finish properly.
- Redundant, missing, or misplaced dependencies.

To mitigate these issues, the following actions will be taken:

- Ensure that all npm commands work properly and follow the same conventions.
- Improve the commands where possible.
- Move correctly the dependencies between packages.

When reviewing, pay special attention to the documentation (the modified READMEs).

#taskid 11601

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,8 +27,8 @@ Mono-repository for the [MathType](http://www.wiris.com/en/mathtype) Web plugins
   - [Publishing](#publishing)
 - [Examples for developers](#examples-for-developers)
 - [More information](#more-information)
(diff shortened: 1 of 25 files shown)


dcanetma:
I've added a little update to fix the table of contents index on both main READMEs.

dcanetma (approved):

PR #252, opened 2021-05-17
KB-11601: Add an ADR document for the shell command wrapper of choice 

dcanetma:
### Context

When writing custom scripts in the node project that perform common tasks like file manipulation (deleting, moving ...) we want to ensure cross-platform compatibility while keeping a simple syntax and reducing vulnerability issues. 

### Decision

We decided to use shx package as a wrapper around ShellJS Unix commands in all scripts of the project.

### DIFF
--- a/docs/adr/003--use-shx-wrapper-for-shell-commands.md
+++ b/docs/adr/003--use-shx-wrapper-for-shell-commands.md
@@ -48,7 +48,7 @@ Using only one library introduces homogeneity and makes development easier.
 
 The main drawback could be overhead: shx contains a lot more functionality than we need by the moment.
 
-We've also added a new root-level thid-party depdendency that will need to be maintained in order to reduce vulnerability risks. 
+We've also added a new root-level thid-party dependency that will need to be maintained in order to reduce vulnerability risks. 
 
 ## Links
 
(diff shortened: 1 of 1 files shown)


docs/adr/003--use-shx-wrapper-for-shell-commands.md
```diff
@@ -0,0 +1,57 @@
...
+It has more than 150K downloads per week, is well-maintaned by the community and it occupies less space than our current past alternatives: rimfaf and copyfiles.
+
+
+## Consequences 
+
+Using only one library introduces homogeneity and makes development easier.
+
+The main drawback could be overhead: shx contains a lot more functionality than we need by the moment.
+
+We've also added a new root-level thid-party depdendency that will need to be maintained in order to reduce vulnerability risks.
```
xaviripo, line 51: depdendency -> dependency

xaviripo (changes requested):

docs/adr/003--use-shx-wrapper-for-shell-commands.md
dcanetma: Fixed

carlawiris (approved):

PR #254, opened 2021-05-18
[Mathtype] "removeImgDataAttributes" function errors when accessing an empty element object

carla-at-wiris:
## Description
There's an error on the mathtype-html-integration-devkit image.js file with a function that deletes the image attributes. This would be okey if previously it checked if the image actually has attributes, in the case it doesn't, throws and access error.

## Steps to reproduce
The client didn't provide any steps to reproduce and we couldn't either, but it's clearly a bug in our code that can be triggered at any time.

## Proposed solution
Solves the error when trying to access and undefined variable on the image.js file in html.integration-devkit package: we check, before acessing the variable attributes, that this actually exists. and has attributes.

This solves [#taskid 12420](https://wiris.kanbanize.com/ctrl_board/2/cards/12420/details/)


xaviripo (approved):

dcanetma (approved):

dcanet-at-wiris:
Merged to stable from another branch.
Added to release v7.27.0

PR #255, opened 2021-05-19
KB-7664: Parse HTML entities in LaTeX

xaviripo:
## Description

This PR solved bug [KB-7664](https://wiris.kanbanize.com/ctrl_board/2/cards/7664/details/) which causes LaTeX rendering to behave strangely with the "<" character and spaces.

The fix included solves a bug that stops the editor from opening when a "<" is included in a LaTeX formula, which is not the same as reported by the client, but was stopping the reproduction of the bug. The bug reported by the client is not reproducible since 7.26.0 and thus can be considered resolved.

## Steps to reproduce

1. Open the editor
2. Type "$$x<1$$" (no spaces)
3. Press "reload" or otherwise render the formula

---

[#taskid 7664](https://wiris.kanbanize.com/ctrl_board/2/cards/7664/details/)


dcanetma (approved):
Ready to merge.

carlawiris (approved):

PR #258, opened 2021-05-20
Add the 'repository' key on all editor packages and set dev-kit dependency version

dcanetma:
### Use '^' for mathtype-html-integration-devkit dependency version

`"@wiris/mathtype-html-integration-devkit": "^1.5.1"
`

### Add the 'repository' key on all editor packages 

The MathType published packages are not properly linked
to GitHub so it's not possible to manage the
third-party dependency graph.

Adding the 'repository' key on all editor packages should
fix this.

Also on this commit:

- Remove unused 'esdoc' settings on generic.
- Set https on homepage setting.

### DIFF
--- a/packages/mathtype-ckeditor4/package.json
+++ b/packages/mathtype-ckeditor4/package.json
@@ -17,12 +17,13 @@
     "mathtype",
     "wiris"
   ],
-  "homepage": "http://www.wiris.com/",
+  "repository": "https://github.com/wiris/html-integrations/tree/stable/packages/mathtype-ckeditor4",
+  "homepage": "https://www.wiris.com/",
   "bugs": {
     "email": "support@wiris.com"
   },
   "license": "MIT",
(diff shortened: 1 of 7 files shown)


xaviripo:
Unfortunately, Lerna automatically removes the `^` from the version field of the devkit dependency when publishing a version, and it sets the latest version instead. This means that either we leave the `^` out for good, or we add it back manually after each release.

dcanetma:
> Unfortunately, Lerna automatically removes the `^` from the version field of the devkit dependency when publishing a version, and it sets the latest version instead. This means that either we leave the `^` out for good, or we add it back manually after each release.

Thanks @xaviripo !
I didn't knew that. :+1: 

I would stick to the first option: leave the ^ out for good and don't bother to change it manually after each release. No biggy.

xaviripo (approved):

PR #262, opened 2021-05-25
Overlapping issue with 'Image' plugin modal dialog window

### LINKED ISSUE #134: When any other dialog box in opened , when mathtype is already opened , either it should close or be underneath the dialog overlay
Steps to reproduce the issue
1.Open wiris editor
2.Open image dialog box or preview box or media
3. The wiris editor comes on top of that.
Attaching images
image
image
image

carla-at-wiris:
## Description

When any other dialog box in opened , when mathtype is already opened , either it should close or be underneath the dialog overlay, but instead it overlaps any other modal dialog.

## Steps to reproduce

1.Open wiris editor
2.Open image dialog box or preview box or media
3. The wiris editor comes on top of that.

## How it's solved

Since tinymce doesn't have a function or event to check if there are active pop ups (mathtype modal dialog, for example), the solution included in this PR disables the editor any time the mathtype dialog it's opened, and enables it again once the dialog closes.

Closes #134

[#taskid 7008](https://wiris.kanbanize.com/ctrl_board/2/cards/7008/details/)


dcanetma:
I've rebased this branch with the last changes on the main one.

Therefore, and in order to use the `CHANGELOG` strategy proposed on the last included changes, I've took the liberty to made these two changes:

- The title of this PR has ben renamed.
- The text has been used as a new entry under the "Unreleased" section of the `CHANGELOG` document.

dcanetma (changes requested):
This is quite an elegant solution. Congratulations. :+1: 

Disabling the TinyMCE toolbox and edit area while the MathType or ChemType editor is open sounds like a coherent behavior.

About the solution on this PR, it switches correctly the read-only mode when the formula is Updated or the Cancel button is pressed.

Unfortunately, it doesn't works as expected if the 'Esc' key is pressed when the MathType editor modal window is open. Then, the image button is disabled and can't be used anymore.

Could you have a look?

It may be worth consider if there's an event on the TinyMCE API that can be used for this. :thinking:

PR #268, opened 2021-05-31
Solve Froala+react render issues

carla-at-wiris:
## Description

When opening the Froala 3 demo on React, the formulas that are written in HTML inside the editor are not rendered, but new formulas can be added and are rendered properly.

## Steps to reproduce

Open a demo on froala3 + react without the changes on this PR, and the formulas inserted through html will not render. With the solution, they should render.

---

[#taskid 11921](https://wiris.kanbanize.com/ctrl_board/2/cards/11921/details/)


dcanetma (changes requested):
Using a `setTimeout` won't be an acceptable solution for this issue.

demos/react/froala3/src/index.js
```diff
@@ -64,22 +63,23 @@ const froalaConfig = {
     initialized() {
       // Get and set the editor and wiris versions in this order.
       Generic.setEditorAndWirisVersion(FroalaEditor.VERSION, pluginVersion);        //eslint-disable-line
-      setTimeout(() => {
-        FroalaEditor.INSTANCES[0].html.set(document.getElementById('transform_content').innerHTML);
-      }, 500);
+
+      // Parse the initial content set on the editor through html to render it
+      const contentRendered = WirisPlugin.Parser.initParse(content);
+      FroalaEditor.INSTANCES[0].html.set(contentRendered);
```
dcanetma, line 69: I'm not sure if this approach would work if there's more than one editor instance on the page.
Have you validate this circumstance?

demos/react/froala3/README.md
```diff
@@ -80,14 +80,13 @@ You will also see any lint errors in the console.
         htmlAllowedTags: ['.*'],
         htmlAllowedAttrs: ['.*'],
         htmlAllowedEmptyTags: ['mprescripts'],
-        imageResize : false,
-        key: 'CA5D-16E3A2E3G1I4A8B8A9B1D2rxycF-7b1C3vyz==',
-        heightMax: 310,
-        useClasses: false
+        // imageResize : false,
+        // heightMax: 310,
+        // useClasses: false
```
dcanetma, line 85: Since these are instructions inside a README file, I would refrain to include commented lines.

demos/react/froala3/README.md
```diff
@@ -70,7 +70,7 @@ You will also see any lint errors in the console.
     // Define the toolbar content and the editor configuration.
     const toolbar = ['wirisEditor', 'wirisChemistry'];
     const froalaConfig = {
-        iframe: true,
+        // iframe: true,
```
dcanetma, line 73: Since these are instructions inside a README file, I would refrain to include commented lines.

At the same time, it's not clear if the iframe value needs to be set to its default value.
Is it different from 'true'?

And I'm curious if this change is responsible for the fix.

dcanetma (changes requested):

demos/react/froala3/README.md
carlawiris: The change is not responsible for the fix, i forgot to comment the line.

demos/react/froala3/README.md
carlawiris: Ok

demos/react/froala3/src/index.js
carlawiris: If there is more than one instance, it'll render just the first one. But since this change is for the code of the integration, it can be changed if you add more instances. Also, lots of users will have their own integration code that can be different from ours, so this is just as an example. Maybe i can add a comment to specify it'll only render the first instance of the page, how does it look like?

xripoll-at-wiris (approved):

PR #277, opened 2021-06-09
Fix Undo & Redo button are broken in rich text toolbar when there is Wiris in content TinyMCE

### LINKED ISSUE #275: Redo button is broken in tinymce rich text toolbar when there is Wiris in content
Replicable in https://demo.wiris.com/integrations/tinymce/
1. make changes to richtext content
2. undo button is enabled
3. click on undo button

**Expected result**
redo button should be enabled
**Actual result**
redo button is still disabled

carla-at-wiris:
## Description

This PR solves an issue with mathtype and TinyMCE (both 4 and 5) that breaks the undo and redo actions.

## Steps to reproduce

1. make changes to richtext content (it doesn't need to involve editing or creating a mt formula)
2. undo button is enabled
3. click on undo button
4. The undo button and redo button don't work any more.

## Why this happens and how is it solved

The lines that breaks the buttons are the 287, 288 & 289:

```js
// Be careful. The next line clear the changes queue
// because is only executed these lines on init.
editor.undoManager.clear();
```

This lines are suposed to be called just once (as the comment says) when the on init function is called, which is theoretically once.

But this oninit function creates a listener that is activated every time a change in the editor is produced, hence the editor.undoManager.clear() is called. This causes the break because clears the queue of undo and redo action storage on every editor change.

The solution resides on deleting the line that clears que queue in this listener and move it to the onInit logic that is actually just called once as we desire.

---

Closes #275 

[#taskid 13094](https://wiris.kanbanize.com/ctrl_board/2/cards/13094/details/)


packages/mathtype-tinymce/editor_plugin.src.js
```diff
@@ -286,7 +286,7 @@ export const currentInstance = null;
                   image.removeAttribute('data-mce-style');
                   // Be careful. The next line clear the changes queue
                   // because is only executed these lines on init.
-                  editor.undoManager.clear();
+                  // editor.undoManager.clear();
```
xripoll-at-wiris, line 289: I'd rather delete the line if you think it's unnecessary than leave it hanging.

xripoll-at-wiris (changes requested):

packages/mathtype-tinymce/editor_plugin.src.js
carlawiris: okey!

xripoll-at-wiris (approved):

dcanet-at-wiris (changes requested):
Changes are fine, but would you be so kind to update the CHANGELOG file accordingly?

Like it's explained here:

> See https://github.com/wiris/integrations-guides/blob/main/changelog/README.md

This way it will be easier to maintain the next release changelog.

At the same time, I'd need you to:

- Squash the commit messages into one, to avoid adding extra lines to the git log.

And I'd suggest to:

- Rewrite the commit message to make it more understanble. For example:

````
Fix "Redo button is broken in TinyMCE toolbar ..." (#275)
````

Thanks

dcanet-at-wiris (approved):
Ready to merge

dcanet-at-wiris:
Merged to stable from another branch.
Released on version v7.27.0

PR #398, opened 2021-09-09
CKEditor5 integration language changes the toolbar buttons language except the wiris ones

carla-at-wiris:
## Description

The mathtype integration with CKEditor5, made both from CKEditor builder or npm dependencies, does not update the tooltip language. The editor and mathtype modal are both in the specified language while the toolbar Wiris buttons are shown in English. 

This PR solves this issue by making the label button property dynamic instead of static.

## Steps to reproduce

1. Open a CK5 demo
2. Define a language following the CKEditor5 guidelines
3. Check the Wiris buttons labels are set in the defined language

---

[#taskid 13890](https://wiris.kanbanize.com/ctrl_board/2/cards/13890/details/)


xripoll-at-wiris (approved):

dcanet-at-wiris (changes requested):
It looks real good, Carla. Thanks!

But I'd need you to review some details some details on the implementation that I would like to discuss together. 😃 

I'm adding comments on the affected lines of code so we can discuss each subject in its thread. 🤓  

Thanks!

dcanet-at-wiris:
**Review: Changes not documented**

I believe you missed to update the Changelog file ([CHANGES.MD](/CHANGES.md)).

We should update the changes done on each package under the [Unreleased] section.

packages/mathtype-ckeditor5/src/plugin.js
```diff
@@ -133,7 +134,7 @@ export default class MathType extends Plugin {
       view.bind('isEnabled').to(editor.commands.get('MathType'), 'isEnabled');
 
       view.set({
-        label: 'Insert a math equation - MathType',
+        label: StringManager.get('insert_math'),
```
dcanet-at-wiris, line 137: That's an excellent solution.

Also, in the case of the CKEditor5 implementation, we may use their message localization support.

Taking advantadge or the

```Javascript

    // The translation function.
    const { t } = editor.locale;

    // The localized label.
    const label = t(StringManager.get('insert_math'));


    buttonView.set( {
        label,
        ...


```

It's quite cool, since even though we could add the localization strings ourlselves, they assume the translation work themselves. 

> _In case of reusing messages, you will not need to worry about translating them as all work will be done by the CKEditor 5 team and Transifex translators. Obviously, your help in translating will still be appreciated!_

See: https://ckeditor.com/docs/ckeditor5/latest/framework/guides/deep-dive/ui/localization.html

dcanet-at-wiris: On a second thought, I think your approach is correct.

What I'm proposing won't work, while the solution you're proposing will work fine in all cases.

Thanks!

packages/mathtype-ckeditor4/plugin.src.js
```diff
@@ -266,15 +267,15 @@ export class CKEditor4Integration extends IntegrationModel {
     init(editor) {
       editor.ui.addButton('ckeditor_wiris_formulaEditor', {
 
-        label: 'Insert a math equation - MathType',
+        label: StringManager.get('insert_math'),
```
dcanet-at-wiris, line 270: I've changed the language on the MathTypeParameters and the tooltip text on the CKEditor4 doesn't seems to change to the right language.

```js
  language: 'de',
  mathTypeParameters: {
    editorParameters: { language: 'de' }, // MathType config, including language
  },
```


Screenshot from 2021-09-10 14-52-55

carla-at-wiris: That is because CKEditor5 works very differently than the other integrations, so the change was easy and straight forward. In the other editors, the buttons labels have been changed with a dynamic strategy that technically takes the editor language to define the button label. Even so, since the specific way those integrations work, the detected language is always English, so we'll implement the necessary changes on a different task.

dcanet-at-wiris (approved):
Ready to merge

PR #449, opened 2021-10-20
Solve error when canceling formula edition

### LINKED ISSUE #445: Error when inserting formula in specific case
## Description

* There is an error when you try to edit and insert existing formula in specific case.

## Environment

https://demo.wiris.com/integrations/generic/

5.27.1 (latest till current date)

Generic HTML integrations
Chrome latest

Windows / Linux

## Steps to reproduce

1. Double click on the fist formula in upper example ((https://demo.wiris.com/integrations/generic/)) 
2. Popup is opened with formula editor
3. Type anything in editor 
3. Click cancel button
4. Alert will be shown with message: "Are you sure you want to leave? The changes you made will be lost."
5. Click cancel
6. Click insert

## Expected result

Edited formula should be inserted

## Actual result

** Callstack:

```
wirisplugin-generic.js:1 Uncaught TypeError: Cannot read property 'attributes' of null
    at Function.value (wirisplugin-generic.js:1)
    at e.value (wirisplugin-generic.js:1)
    at e.value (wirisplugin-generic.js:1)
    at a.value (wirisplugin-generic.js:1)
    at a.value (wirisplugin-generic.js:1)
    at a.value (wirisplugin-generic.js:1)
    at e.value (wirisplugin-generic.js:1)
    at e.value (wirisplugin-generic.js:1)
```

*What is the actual outcome of the steps above?*

Formula is not inserted and error is thrown in dev console

carla-at-wiris:
## Description

On any editor of the ones supported by wiris, there is an error when you try to edit and insert existing formula in the case where you click the cancel button before inserting the formula.
The error appears on the console and the formula is not inserted. 

This error was caused due to a function that was getting called in this specific situation, detailed above, when it should onnly be called when we want to close the editing modal.

This PR solves the issue by changing the places where the mentioned function gets called: Only when the modal window is closed through a close button.

## Steps to reproduce

1. Open any demo. (The client used the Generic one)
2. Double click on an existing formula.
3. Type anything.
4. Click cancel button
5. Alert will be shown with message: "Are you sure you want to leave? The changes you made will be lost."
6. Click cancel.
7. Click insert.
8. The console error appears, the formula is not inserted and the modal is not closed

---

Closes #445 

[#taskid 16499](https://wiris.kanbanize.com/ctrl_board/2/cards/16499/details/)


dcanet-at-wiris (approved):
Good work, Carla!
I've been able to validate the solution on both generic and froala editors.

I would recommend to rebase all commit messages to only one before merging the branch to `KB-15953-BUGSPRINT`.

On the other hand I've found two issues that I've decided to address myself. Let me explain below, so you can share your thoughts, too.

### Fix the 'demo-html5-generic': chokidar

The  'demo-html5-generic' wasn't working on development mode.

This error appeared on console:

```bash
Error: No version of chokidar is available. Tried chokidar@2 and chokidar@3.'
chokidar@3: Error: Cannot find module 'chokidar

```
To fix this, chokidar needed to be installed manually.

### Fix vulnerabilities with an audit and remove jest

The 'jest' dependency was still present at the root of the project.

This library was supposed to be removed from the project before.
This change fix this.

PR #479, opened 2021-12-01
KB-17724: Add an ADR document for XSS vulnerability when browser engines parsing HTML

icaparros-at-wiris:
## Description

Detected an XSS vulnerability on the Mathtype devkit package. A user could inject malicious to a website code encaplusating them inside a <math> tag.

In order to prevent XSS attacks it is proposed to use the DOMPurify library. An open-source XSS sanitizer for HTML, MathML and SVG.

## Decision

DOMPurify sanitize HTML and prevents XSS attacks. You can feed DOMPurify with string full of dirty HTML and it will return a string removing the dangerous HTML.

DOMPurify will be added to @wiris/mathtype-html-integration-devkit so this dependency will affect to all the html-integrations packages.

---

[#taskid 17724](https://wiris.kanbanize.com/ctrl_board/2/cards/17724/details/)

### DIFF
--- a/docs/adr/005-Use-HTML-sanitizer-to-avoid-XSS-attacks.md
+++ b/docs/adr/005-Use-HTML-sanitizer-to-avoid-XSS-attacks.md
@@ -1,0 +1,82 @@
+# 005. Use DOMPurify as HTML sanitizer to avoid XSS attacks
+
+Date: 29-no-2021
+
+## Status
+
+ACCEPTED
+
+## Summary
+
+_An XSS vulnerability has been detected on the MathType devkit package._
+_A user could inject malicious code using the editor by encapsulating them inside a `<math>` tag._
+_In order to prevent XSS attacks, we will use the DOMPurify third-party library. An open-source XSS sanitizer for HTML, MathML and SVG code._
+
+## Context
+
+All browser engines, during _parsing_ step, they turn the data received over the network into the DOM. Given the following block, which uses the `<math>` tag:
+
+```html
+<math>
+  <style>
+    <img src onerror="alert(1)" />
+  </style
+</math>
+```
+
+The browser engine will generate it to:
+
+```html
+<MathML math>
+  <MathML style> <MathML img src onerror="alert(1)" /></MathML
+></MathML>
+```
+
+Normally, an `<img>` inside an `<style>` will be parsed as a `#text` inside the `<MathML style>` tag. Nevertheless, this behavior changes when it's encapsulated inside a `<math>` tag, and interprets the `img` as a `<MathML>` independent block. 
+
+As a result, the HTML will be unwrapped and written on the DOM following this format:
+
+```html
+<math>
+  <img src onerror="alert(1)" />
+  <style></style>
+</math>
+```
+
+This parsing behavior can be used to inject malicious JavaScript on a website. In order to avoid this vulnerability, is needed to filter all the HTML added from the editor.
+
+Implementing a sanitize system can be laborious, and it will be difficult to maintenance in order to detect as much XSS vulnerabilities as possible.
+
+## Decision
+
+DOMPurify sanitize HTML and prevents XSS attacks. You can feed DOMPurify with strings full of dirty HTML, and it will return a string without ant dangerous HTML.
+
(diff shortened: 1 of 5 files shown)


docs/adr/005-Use-HTML-sanitizer-to-avoid-XSS-attacks.md
```diff
@@ -0,0 +1,83 @@
...
+
+Date: 29-no-2021
+
+## Status
+
+PROPOSED  
+
+## Summary
+
+*CKEditor4 team detected an XSS vulnerability on the Mathtype devkit package. A user could inject malicious to a website code encaplusating them inside a <math> tag.*
```
carla-at-wiris, line 11: I would omit that the CKEditor4 team found the vulnerability, and add that we have a XSS vulnerability since we don't review if the code obtained is correct, or something similar

docs/adr/005-Use-HTML-sanitizer-to-avoid-XSS-attacks.md
```diff
@@ -0,0 +1,83 @@
...
+
+DOMPurify sanitize HTML and prevents XSS attacks. You can feed DOMPurify with string full of dirty HTML and it will return a string removing the dangerous HTML.
+
+DOMPurify will be added to `@wiris/mathtype-html-integration-devkit` so this dependency will affect to all the `html-integrations` packages.
+
+### Pros and Cons of the Options 
+
+#### Implement our own Javascript library for that
+
+- Bad, because it is a huge task to detect all of possible XSS injections.
```
carla-at-wiris, line 58: I thing that there's a good part on it, since we would have full control on the library. By adding this good clause, i don't mean that the option to create our own library is better, but it's good to have documented all pros and cons

carla-at-wiris (commented):
Apart from the two comments i added, the ADR looks fine

dcanet-at-wiris (approved):

dcanet-at-wiris (changes requested):
I believe you forgot to update the CHANGELOG file.

A part from tjat, ready to merge.

xripoll-at-wiris (approved):
Very good ADR. Watch out for typos (egine) and style (e.g the capitalization of MathType).

docs/adr/005-Use-HTML-sanitizer-to-avoid-XSS-attacks.md
icaparros-at-wiris: It's makes sense, I have modified this part.

docs/adr/005-Use-HTML-sanitizer-to-avoid-XSS-attacks.md
icaparros-at-wiris: Added the good part.

carla-at-wiris (approved):

dcanet-at-wiris (changes requested):
## Review

Nice job, Isaac. Congrats.

The code you've added looks good and it passes the lint validation.

I'm glad you've decided to create an ADR document. It is very well-written and specific. Kudos to you and @carla-at-wiris .

At the same time, it seems like the right decission. The newly added library, [DOMPurify](https://github.com/cure53/DOMPurify), seems like the right choice. ( 0 vulnerabilites added to the package, also). 

> BTW, I take the chance to change some text at the end, as we discussed during the pair-programming session.  
> Feel free to rebase the commit before merging the branch.

## Conclusion: changes required.

On the other hand, I am afraid we all missed something. The [README.md](/packages/mathtype-html-integration-devkit/README.md) file for the affected package includes a section named "Dependencies".

It details all the dependencies, not the dev ones, added to the package. 

See: https://github.com/wiris/html-integrations/tree/KB-17724/packages/mathtype-html-integration-devkit#dependencies

Imho, we should add the new [dompurify](https://github.com/cure53/DOMPurify) to that section.

dcanet-at-wiris (approved):
Ready to merge.

Just to make sure, though.

Is there no other functions in the SDK, other than `codeImgTransform`, that need to use the `DOMPurify.sanitize` function, right?

I believe you mentioned on our last daily meeting but I'm not sure if I am right.

PR #480, opened 2021-12-02
Solve parser functions not working on Generic Integrations

carla-at-wiris:
## Description

This PR solves a bug with the Parser functions and the Generic Integration. The bug turned off the Parser functions, so they returned the exact same parameter they received, instead of transforming it. For example, executing the following function:
```js
WirisPlugin.Parser.initParse('<math><mo>x</mo></math>');
```
should return the mathml transformed into an image, but instead returns the same mathml.

This behavior is caused due to a bad configuration variable on the `global.js` mathtype-generic package file, and a wrong way to initialize the generic editor instance on the HTML5, Angular and React demos.

## Steps to reproduce

1. Open the Generic demo.
2. With F12, open the web inspector.
3. On the console tab, write the following code: `WirisPlugin.Parser.initParse('<math><mo>x</mo></math>');`
4. See that the returned object is the same mathml instead of a formula image.

---

[#taskid 16836](https://wiris.kanbanize.com/ctrl_board/2/cards/16836/details/)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,6 +18,11 @@ Last release of this project is was 30th of September 2021.
   - Added missing 'www' on wiris.net documentation links.
   - Destroy events from CKEditor5Integration when CK5 is destroyed.
 
+- Solve Parser not working on Generic Integrations (Issue - #450)
+
+  - Modify the generic package to use properly the parser functions.
+  - Modify the demo to initialyze the editor exposing it to the window, so it can have the necessary configurations to use the Parser class.
+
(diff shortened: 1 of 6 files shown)


xripoll-at-wiris (approved):

dcanet-at-wiris (approved):

dcanet-at-wiris:
@carla-at-wiris Be careful not to merge these changes to the wrong target branch

icaparros-at-wiris (approved):

PR #484, opened 2021-12-07
Fix Froala v3 & v4 + Angular treating formulas as images

carla-at-wiris:
## Description

The Wiris formulas images should have a different behavior than regular images when clicking into them, meaning that the image toolbar for formulas should include the mathtype buttons when specified. With angular framework and Froala v3 or v4 editor, this behavior was not working, instead, the formulas were treated as regular images even when the image toolbar had specified the mathtype buttons.

The bug was solved by:
- Deleting jQuery residual code on mathtype-froala3 package, since Froala3 and above works with pure JavaScript.
- Migrating webpack v4 to v5.

This PR also includes:
- Removing jQuery dependency from html5+froala demo.
- Updating Froala version on angular and html5 + froala demos.

## Steps to reproduce

1. Open the angular + froala demo.
2. Click on any formula on the editor.
3. Check that the image toolbar doesn't even open the image toolbar and an error appears on the console.

---

[#taskid 14726](https://wiris.kanbanize.com/ctrl_board/2/cards/14726/details/)

### DIFF
--- a/packages/mathtype-froala3/wiris.src.js
+++ b/packages/mathtype-froala3/wiris.src.js
@@ -1,5 +1,3 @@
-/* globals $ */
-
 import IntegrationModel from '@wiris/mathtype-html-integration-devkit/src/integrationmodel';
 import Configuration from '@wiris/mathtype-html-integration-devkit/src/configuration';
 import Parser from '@wiris/mathtype-html-integration-devkit/src/parser';
@@ -279,7 +277,7 @@ export class FroalaIntegration extends IntegrationModel {
       currentFroalaIntegrationInstance.hidePopups();
       currentFroalaIntegrationInstance.core.getCustomEditors().disable();
       const imageObject = currentFroalaIntegrationInstance.editorObject.image.get();
-      if (typeof imageObject !== 'undefined' && imageObject !== null && imageObject.hasClass(WirisPlugin.Configuration.get('imageClassName'))) {
+      if (typeof imageObject !== 'undefined' && imageObject !== null && imageObject[0].classList.contains(WirisPlugin.Configuration.get('imageClassName'))) {
         currentFroalaIntegrationInstance.core.editionProperties.temporalImage = imageObject[0];
         currentFroalaIntegrationInstance.openExistingFormulaEditor();
       } else {
@@ -294,19 +292,19 @@ export class FroalaIntegration extends IntegrationModel {
     // Value can be undefined.
     if (selectedImage) {
       if (($btn.parent()[0].hasAttribute('class') && $btn.parent()[0].getAttribute('class').indexOf('fr-buttons') === -1) || (selectedImage[0]
-                && ($(selectedImage[0]).hasClass(Configuration.get('imageClassName')) || $(selectedImage[0]).contents().hasClass(Configuration.get('imageClassName'))))) { // Is a MathType image.
-        // Show MathType icons if previously were hiden.
+                && (selectedImage[0].classList.contains(Configuration.get('imageClassName')) || selectedImage[0].contents().classList.contains(Configuration.get('imageClassName'))))) { // Is a MathType image.
+        // Show MathType icons if previously were hidden.
         $btn.removeClass('fr-hidden');
         // Disable resize box.
-        if (!$('#wrs_style').get(0)) { // eslint-disable-line no-undef
-          $('head').append('<style id="wrs_style">.fr-image-resizer {pointer-events: none;}</style>');
+        if (!document.getElementById('wrs_style')) { // eslint-disable-line no-undef
+          document.getElementsByTagName('head')[0].append('<style id="wrs_style">.fr-image-resizer {pointer-events: none;}</style>');
         }
       } else { // Is a non-MathType image.
         // Hide MathType icons.
         $btn.addClass('fr-hidden');
         // Enable resize box (if it was configured).
-        if ($('#wrs_style').get(0)) {
-          $('#wrs_style').get(0).remove();
+        if (document.getElementById('wrs_style')) {
+          document.getElementById('wrs_style').remove();
         }
       }
     }
@@ -326,7 +324,7 @@ export class FroalaIntegration extends IntegrationModel {
       currentFroalaIntegrationInstance.hidePopups();
       currentFroalaIntegrationInstance.core.getCustomEditors().enable('chemistry');
       const imageObject = currentFroalaIntegrationInstance.editorObject.image.get();
-      if (typeof imageObject !== 'undefined' && imageObject !== null && imageObject.hasClass(WirisPlugin.Configuration.get('imageClassName'))) {
+      if (typeof imageObject !== 'undefined' && imageObject !== null && imageObject[0].classList.contains(WirisPlugin.Configuration.get('imageClassName'))) {
         currentFroalaIntegrationInstance.core.editionProperties.temporalImage = imageObject[0];
         currentFroalaIntegrationInstance.openExistingFormulaEditor();
       } else {
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -29,6 +29,9 @@ Last release of this project is was 30th of September 2021.
   Latex formulas have a semantics tag that requires its inside mathml to be inside a `mrow` tag.
   Added this tag on the Latex formula generation.
 
+- Fix Angular + Froala (v3 & v4) treting Wiris formulas as images.
+  - Update Webpack to V5 and remove jQuery, on mathtype-froala3 and its demos.
+
 ## 7.27.2 - 2021-11-26
 
 ## CKEditor5 filtering mechanism
--- a/demos/angular/froala/package.json
+++ b/demos/angular/froala/package.json
@@ -13,7 +13,7 @@
   "dependencies": {
     "@wiris/mathtype-froala3": "^7.27.0",
     "@wiris/mathtype-html-integration-devkit": "*",
-    "angular-froala-wysiwyg": "^4.0.4"
+    "angular-froala-wysiwyg": "^4.0.7"
   },
   "devDependencies": {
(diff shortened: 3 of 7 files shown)


packages/mathtype-froala3/wiris.src.js
```diff
@@ -294,19 +292,19 @@ export class FroalaIntegration extends IntegrationModel {
     // Value can be undefined.
     if (selectedImage) {
       if (($btn.parent()[0].hasAttribute('class') && $btn.parent()[0].getAttribute('class').indexOf('fr-buttons') === -1) || (selectedImage[0]
-                && ($(selectedImage[0]).hasClass(Configuration.get('imageClassName')) || $(selectedImage[0]).contents().hasClass(Configuration.get('imageClassName'))))) { // Is a MathType image.
+                && ((selectedImage[0]).classList.contains(Configuration.get('imageClassName')) || (selectedImage[0]).contents().classList.contains(Configuration.get('imageClassName'))))) { // Is a MathType image.
```
xripoll-at-wiris, line 295: Maybe the parentheses around `selectedImage[0]` can be dropped? And same thing with the parentheses around the whole line.
In fact, if you think it would be more readable, you could set up variables to break down the line into smaller chunks, e.g. a variable for `selectedImage[0]`, or a variable for each of the three conditionals joined by `&&`.

packages/mathtype-froala3/wiris.src.js
```diff
@@ -294,19 +292,19 @@ export class FroalaIntegration extends IntegrationModel {
...
         // Show MathType icons if previously were hiden.
         $btn.removeClass('fr-hidden');
         // Disable resize box.
-        if (!$('#wrs_style').get(0)) { // eslint-disable-line no-undef
-          $('head').append('<style id="wrs_style">.fr-image-resizer {pointer-events: none;}</style>');
+        if (!document.getElementById('wrs_style')) { // eslint-disable-line no-undef
+          (document.getElementsByTagName('head')[0]).append('<style id="wrs_style">.fr-image-resizer {pointer-events: none;}</style>');
```
xripoll-at-wiris, line 300: Same thing about the parentheses, check if they can be dropped for better readability.

xripoll-at-wiris (approved):

xripoll-at-wiris:
I recommended rewriting some lines to take the chance to make them more readable.

Good to see unnecessary dependencies being dropped though.

dcanet-at-wiris (changes requested):
Good job, Carla!
Removing the jQuery dependency from the Froala package is such a good move. 
I am happy that you decide to get rid of this dependency. 🥳 

The reason I am asking for review is that I believe that there are other uses of the jQuery's `hasClass` function on other parts of the `mathtype-froala3` plugin.

Like on these next lines:

- packages/mathtype-froala3/wiris.src.js:280
- packages/mathtype-froala3/wiris.src.js:327

I believe you should remove this function calls to in order to avoid errors.

At the same time, I've found that the `mathtype-froala` plugin also holds references to this same jQuery function.

- packages/mathtype-froala/wiris.src.js:279
- packages/mathtype-froala/wiris.src.js:295
- packages/mathtype-froala/wiris.src.js:327


Since the current installation instructions don't mention that the jQuery dependency needs to be added manually and it's not included on the package dependency, I guess this could be causing unexpected behaviors in some use cases.

Would you mind to have a look to it, on this package, too?

packages/mathtype-froala3/wiris.src.js
carla-at-wiris: settings variables for selectedImage[0] could lead to errors, since the if clause verifies the existence of that selectedImage[0], but for the parentheses parts you're right

packages/mathtype-froala3/wiris.src.js
carla-at-wiris: > Good job, Carla! Removing the jQuery dependency from the Froala package is such a good move. I am happy that you decide to get rid of this dependency. partying_face
> 
> The reason I am asking for review is that I believe that there are other uses of the jQuery's `hasClass` function on other parts of the `mathtype-froala3` plugin.
> 
> Like on these next lines:
> 
> * packages/mathtype-froala3/wiris.src.js:280
> * packages/mathtype-froala3/wiris.src.js:327
> 
> I believe you should remove this function calls to in order to avoid errors.

You're right, i've changed them and tested they work properly

> At the same time, I've found that the `mathtype-froala` plugin also holds references to this same jQuery function.
> 
> * packages/mathtype-froala/wiris.src.js:279
> * packages/mathtype-froala/wiris.src.js:295
> * packages/mathtype-froala/wiris.src.js:327
> 
> Since the current installation instructions don't mention that the jQuery dependency needs to be added manually and it's not included on the package dependency, I guess this could be causing unexpected behaviors in some use cases.
> 
> Would you mind to have a look to it, on this package, too?

Regarding the Froala 2 package, froala 2 has jQuery as a dependency, the froala team removed this when updating to V3 and above, so jQuery dependencies and functions on this package are necessary

@dcanet-at-wiris

carla-at-wiris:
> Good job, Carla! Removing the jQuery dependency from the Froala package is such a good move. I am happy that you decide to get rid of this dependency. partying_face
> 
> The reason I am asking for review is that I believe that there are other uses of the jQuery's `hasClass` function on other parts of the `mathtype-froala3` plugin.
> 
> Like on these next lines:
> 
> * packages/mathtype-froala3/wiris.src.js:280
> * packages/mathtype-froala3/wiris.src.js:327
> 
> I believe you should remove this function calls to in order to avoid errors.

You're right, i've changed them and tested they work properly

> At the same time, I've found that the `mathtype-froala` plugin also holds references to this same jQuery function.
> 
> * packages/mathtype-froala/wiris.src.js:279
> * packages/mathtype-froala/wiris.src.js:295
> * packages/mathtype-froala/wiris.src.js:327
> 
> Since the current installation instructions don't mention that the jQuery dependency needs to be added manually and it's not included on the package dependency, I guess this could be causing unexpected behaviors in some use cases.
> 
> Would you mind to have a look to it, on this package, too?

Regarding the Froala 2 package, froala 2 has jQuery as a dependency, the froala team removed this when updating to V3 and above, so jQuery dependencies and functions on this package are necessary

@dcanet-at-wiris

packages/mathtype-froala3/wiris.src.js
carla-at-wiris: Right! I'm on it

dcanet-at-wiris:
> > Good job, Carla! Removing the jQuery dependency from the Froala package is such a good move. I am happy that you decide to get rid of this dependency. partying_face
> > The reason I am asking for review is that I believe that there are other uses of the jQuery's `hasClass` function on other parts of the `mathtype-froala3` plugin.
> > Like on these next lines:
> > 
> > * packages/mathtype-froala3/wiris.src.js:280
> > * packages/mathtype-froala3/wiris.src.js:327
> > 
> > I believe you should remove this function calls to in order to avoid errors.
> 
> You're right, i've changed them and tested they work properly
> 
> > At the same time, I've found that the `mathtype-froala` plugin also holds references to this same jQuery function.
> > 
> > * packages/mathtype-froala/wiris.src.js:279
> > * packages/mathtype-froala/wiris.src.js:295
> > * packages/mathtype-froala/wiris.src.js:327
> > 
> > Since the current installation instructions don't mention that the jQuery dependency needs to be added manually and it's not included on the package dependency, I guess this could be causing unexpected behaviors in some use cases.
> > Would you mind to have a look to it, on this package, too?
> 
> Regarding the Froala 2 package, froala 2 has jQuery as a dependency, the froala team removed this when updating to V3 and above, so jQuery dependencies and functions on this package are necessary
> 
> @dcanet-at-wiris

I don't understand the Froala2 part.
Shouldn't we include the jQuery dependency to the package, then?

dcanet-at-wiris (changes requested):
Froala 2 comment

carla-at-wiris:
> I don't understand the Froala2 part.
> Shouldn't we include the jQuery dependency to the package, then?

It's like Froala-editor dependency, since they're dependencies required for the demo to work, they don't need to be on the package itself

carla-at-wiris:
[Here's the documentation](https://froala.com/wysiwyg-editor/docs/migrate-from-v2/) regarding the migration from Froala V2 to V3, as you can see at the beginning of the post, they mention that in the V3 and above they got rid of the jQuery dependency from V2.
There's also an integration example for the froala-editor V2 and V3, and you can see that they use jQuery on v2.

The npm documentation[ on this page](https://www.npmjs.com/package/froala-editor/v/2.9.7) also shows the steps to follow to start a froala-editor on v2, and they mention that jQuery is a required dependency. 

For that reason, jQuery is necessary for the mathtype-froala2 package, but it's not required to be a package dependency since it must be a demo dependency.
@dcanet-at-wiris

packages/mathtype-froala3/wiris.src.js
```diff
@@ -294,19 +292,19 @@ export class FroalaIntegration extends IntegrationModel {
     // Value can be undefined.
     if (selectedImage) {
       if (($btn.parent()[0].hasAttribute('class') && $btn.parent()[0].getAttribute('class').indexOf('fr-buttons') === -1) || (selectedImage[0]
-                && ($(selectedImage[0]).hasClass(Configuration.get('imageClassName')) || $(selectedImage[0]).contents().hasClass(Configuration.get('imageClassName'))))) { // Is a MathType image.
+                && (selectedImage[0].classList.contains(Configuration.get('imageClassName')) || selectedImage[0].contents().classList.contains(Configuration.get('imageClassName'))))) { // Is a MathType image.
         // Show MathType icons if previously were hiden.
```
carla-at-wiris, line 296: Solve typo

icaparros-at-wiris (approved):

dcanet-at-wiris (approved):
Good job, Carla.
Ready to merge.

PR #496, opened 2021-12-22
[KB-17161] Automated release for npm packages

xripoll-at-wiris:
## Description

This PR adds a new GitHub Actions workflow capable of automatically publishing the front-end npm packages.

It has two main additions:

- ADRs, justifying the technical decisions taken. Please read those and comment if there is any disagreements.
- Code changes, including:
  - A new workflow (`.github/workflows/publish.yml`).
  - A new Expect script (`scripts/publish.exp`) in charge of managing Lerna, used by the previous workflow.

---

[#taskid 17161](https://wiris.kanbanize.com/ctrl_board/2/cards/17161/details/)

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -35,11 +35,6 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
-      # - name: Merge master to stable
-      #   run: |
-      #     git checkout stable
-      #     git merge master
-
       - name: Install expect
         run: sudo apt-get install -y expect
 
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,71 @@
+# Publishes the MathType integrations front-end packages to npm
+#
+# Jobs:
+# - Publish
+#
+name: Publish
+
+# Controls when the action will run
+on:
+  # Ideally we would want this to run on every PR, but GitHub Actions doesn't offer a way to set default values for the
+  # input arguments of the execution.
+  # pull_request:
+  workflow_dispatch:
+    inputs:
+      publish_npm:
+        description: |
+          Whether to publish the frontend packages to npm.
+          Type `publish_npm` to do publish to npm, any other value otherwise.
+          WARNING: This action cannot be undone.
+        required: false
+        default: ""
+      versions:
+        description: |
+          Space-separated list of versions to publish, e.g.: "@wiris/mathtype-ckeditor5=7.27.0 @wiris/mathtype-html-integration-devkit=1.6.1"
+          Packages not mentioned in the list will stay at their current version and will not be published.
+        required: false
+        default: ""
+
+# Array of jobs to run in this workflow
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      # - name: Merge master to stable
+      #   run: |
+      #     git checkout stable
+      #     git merge master
+
+      - name: Install expect
+        run: sudo apt-get install -y expect
+
+      # For safety reasons, by default all npm publishing actions in this pipeline will be done with a dummy private npm registry underneath.
+      # This will only be turned off when the publish_npm flag is set to "publish_npm".
+      - name: Set up a private npm registry and publish
+        run: |
+
+          # In case we want to just emulate the publishing, we set up Verdaccio
+          if [ ${{ github.event.inputs.publish_npm == 'publish_npm' }} ] ; then
+
+            # Add the token to publish to the npm registry
+            npm set "//registry.npmjs.org/:_authToken" ${{ secrets.NPM_TOKEN }}
+
+          else
+
+            # We use Verdaccio to run a local private npm server
+            npm install -g verdaccio
+
+            # Set npm to publish to local registry
+            cp resources/publish/.npmrc ~/.npmrc
+
+            # Run and block verdaccio until it is ready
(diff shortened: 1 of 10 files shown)


icaparros-at-wiris (commented):
Awesome ADRs.

Few days ago, I merged the ADR `005-Use-HTML-sanitizer-to-avoid-XSS-attacks.md` on the `KB-15953-BUGSPRINT` branch. We should rename this ADR to 006-Use-DOMPurify...
or rename your both ADR prefix to 006 and 007.

What It makes more sense?

resources/publish/htpasswd
```diff
@@ -0,0 +1 @@
+username:$apr1$gwQzvrYX$hLkvErWORju3C1RS2.QrW0
```
carla-at-wiris, line 1: is the password for the user name stored as private in some place?

docs/adr/006-use-expect-for-automating-tui-interaction.md
```diff
@@ -0,0 +1,68 @@
...
+
+Shell scripting is a very powerful, but dangerous, solution.
+
+- Good, because it doesn't any dependencies.
+- Bad, because it is very hard to write robust scripts.
+- Bad, because it is complicated to interact with the output of the TUI to automate.
+
+#### Expect
+
+CNPM stands for "Company npm" and is a private npm registry software.
```
dcanet-at-wiris, line 41: Typo alert.
I guess the right text for this paragraph would be:
> "_Expect is a tool for automating interactive applications such as telnet, ftp, passwd, fsck, rlogin, tip, etc._"
> From [Expect website](https://core.tcl-lang.org/expect/index)

dcanet-at-wiris (changes requested):

dcanet-at-wiris:
I've added "Expect Language Support" as a recommended extension to the project.
Since we are adding this extension to our current tool set, I think it's a good idea to add it here, too, for other contributors.

docs/adr/006-use-expect-for-automating-tui-interaction.md
dcanet-at-wiris: I fixed myself, @xripoll-at-wiris

dcanet-at-wiris:
> Awesome ADRs.
> 
> Few days ago, I merged the ADR `005-Use-HTML-sanitizer-to-avoid-XSS-attacks.md` on the `KB-15953-BUGSPRINT` branch. We should rename this ADR to 006-Use-DOMPurify... or rename your both ADR prefix to 006 and 007.
> 
> What It makes more sense?

Thanks for mentioning it @icaparros-at-wiris.

In order to help things out, I've renamed the Verdaccio's ADR myself to number 007, so we avoid the coincidence with #479.

Thanks!

dcanet-at-wiris:
**Important update**.

This PR's branch has been rebased with the `KB-15953-BUGSPRINT` branch.

Update you local branches, accordingly, please.

Sorry for any incovenience.

dcanet-at-wiris (commented):
## Adding `.npmrc` file to the repository 🙀 

I have a question regarding the `.npmrc` file. I am not quite sure if adding this file to the repo is a good move.

If I am not wrong, the default decission would be to include this file to `.gitignore` so it's not added to the repository at all.

I understand that the Verdaccio configuration needs to be included on the repository, in order to work, though.

I am reading the [Verdaccio's documentation site](https://verdaccio.org/docs/authentication/) and I guess one could argue that the user, (developer), could follow some instructions themselves and/or automate the `.npmrc` creation somehow.

```bash
# Add the registry key:
npm adduser --registry http://localhost:4873
# More stuff...
```

What's your take?

Or does the fact that it's not stored on the root folder, and it's more a boilerplate file for an effective CI/CD workflow configuration, that made you decide this option?

resources/publish/htpasswd
dcanet-at-wiris: I'd be more interested to know which command you've used to generate the username token for "username". 😄 

Something like:

```bash
echo -n 'username' | openssl base64
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,71 @@
...
+# Array of jobs to run in this workflow
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      # - name: Merge master to stable
```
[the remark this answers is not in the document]
xripoll-at-wiris: No, I forgot to remove them. Thanks.

resources/publish/htpasswd
xripoll-at-wiris: Good note, I'll add a comment.

> is the password for the user name stored as private in some place?

This line is a pair of the form `${username}:${hashed password}`.
It is unsafe to upload your personal `.htpasswd`, but this username:password combination is just literally `username:password`. This file is used to authenticate in the local Verdaccio instance, it's not used in an actual server or in production.

> I'd be more interested to know which command you've used to generate the username token for "username". smile

Also very good note. The command is:

```sh
htpasswd -c htpasswd username
```

Which prompts for a password. Note that the output will be (probably) be different every time you generate this file, even if you enter the same password.

xripoll-at-wiris:
> Or does the fact that it's not stored on the root folder, and it's more a boilerplate file for an effective CI/CD workflow configuration, that made you decide this option?

Effectively this. The intention is not to override user settings, just to use this file when called on purpose. This shouldn't affect anything else. I think it's wise to store configurations outside of code, e.g. in this case in a separate config file.

dcanet-at-wiris (approved):

PR #498, opened 2021-12-28
Run tests in development mode on local and CI/CD machine

carla-at-wiris:
## Description

The main goals of this task are to be able to run the current set of tests using the node local/build packages, both on development and CI/CD environments. 

This PR includes an improvement on the build command used to launch the demo for testing purposes: it has now turned into two new commands even more specific, indicating the demo they build and the environment they run on:
* build-generic-demo:public
  Builds the generic demo with the published MathType npm packages. 

* build-generic-demo:local
  Build the generic demo using the local MathType npm packages

> The commands are not merged into the run ones because this will add a lot of complexity and it'll turn out as 12 npm scripts or 6 tangled ones. 

### CI implementation improvement
* A new parameter on workflow dispatch: `packages-code`:

  The parameter only accepts the value 'local', on this iteration, to run the tests with the local packages.
  > The parameter MUST be introduced as single quote in order to be used on the jobs.

* A new job called `define-env`:

  Contains a set of basic steps that, create an environment variable, retrieve the event that activated the workflow, assign to the variable the 'local' string depending on the event, and expose it as a job output, so that the testing job can use this as a row on the matrix strategy.
  > The job is prepared to accept the 'public' parameter that will be added in future iterations, for that reason, there're more than one step with the same output.

* A new step on the testing job to install Lerna dependencies when running tests in local following the matrix strategy.

* Updated the testing step, build section with the new build commands.

* Fix some typos.

* Rename the Workflow to a more accurate one, since it doesn't only run tests with published packages anymore.

### Also included on this PR
* Updated the `CHANGELOG.md`
* Updated the testing `README` to include the changes to the commands and the information to run the tests in development mode on the local machine.

You can see the workflow runs with the local packages on the CI/CD machine on the actions tab, the workflow is called `Run Cypress tests with local packages`

## Steps to reproduce

You can run the tests with the local packages both on

**CI/CD environment**:
* On the Actions tab, find the "Run Cypress tests with local packages" workflow.
* Use the workflow dispatch option and set the environment Value as 'local', which is the default.
* Tests will run with the local packages.
> This option is currently disabled since it is required a dummy test on stable branch to enable it, but you can still check the already executed workflows.

**Local machine**:
* Follow the instructions on the docs/development/testing/README.md to run the tests with the local packages.
* Make some changes on the mathtype-generic or mathtype-html-integration-devkit packages to try to make the test fail and verify that the tests run with the local packages.

---

[#taskid 17160](https://wiris.kanbanize.com/ctrl_board/2/cards/17160/details/)


xripoll-at-wiris (approved):
Very good job!

icaparros-at-wiris (approved):
Great job

PR #583, opened 2022-05-27
fix: open MT/CT Editor in Keybord mode when using iOS

icaparros-at-wiris:
## Description

When the user opens the MT/CT Editor using an iOS device with zoomed webpage. The user will not be able to zoom out on the handwrite area.

Now, the MT/CT Editor will be open on keyboard mode when using an iOS device by default.

If the user use Safari and doesn't zoom, the page will be opened in handwrite mode. This will not happen in Chrome, as I didn't found a way to detect the zoom level of the page.

## Steps to reproduce

1. Open a demo on `html-integrations`
2. With an iOS device, navigate to the opened demo using Safari
3. Zoom the web page
4. Open the MT Editor

The MT Editor will be opened in Keyboard mode

5. Close the MT Editor
6. Reset the page
7. Open the MT Editor without zooming the page

The MT Editor will be opened in Handwrite mode

---

[#taskid 24192](https://wiris.kanbanize.com/ctrl_board/2/cards/24192/details/)

### DIFF
--- a/packages/mathtype-html-integration-devkit/src/contentmanager.js
+++ b/packages/mathtype-html-integration-devkit/src/contentmanager.js
@@ -447,6 +447,28 @@ export default class ContentManager {
     }
     this.updateToolbar();
     this.onFocus();
+    
+    if (this.deviceProperties.isIOS) {
+      const zoom = document.documentElement.clientWidth / window.innerWidth;
+
+      if (zoom != 1) {
+        // Open editor in Keyboard mode if user use iOS, Safari and page is zoomed.
+        this.setKeyboardMode();
+      }
+    }
+  }
+
+  /**
+   * Change Editor in keyboard mode when is loaded
+   */
+  setKeyboardMode() {
+    const wrsEditor = document.getElementsByClassName('wrs_handOpen wrs_disablePalette')[0];
+    if (wrsEditor) {
+      wrsEditor.classList.remove('wrs_handOpen');
+      wrsEditor.classList.remove('wrs_disablePalette');
+    } else {
+      setTimeout(ContentManager.prototype.setKeyboardMode.bind(this), 100);
+    }
   }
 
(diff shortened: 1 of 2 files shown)


packages/mathtype-html-integration-devkit/src/contentmanager.js
```diff
@@ -447,6 +479,33 @@ export default class ContentManager {
     }
     this.updateToolbar();
     this.onFocus();
+
+    if (this.deviceProperties.isIOS) {
+      if (this.browserProperties.isSafari) {
+        const zoom = document.documentElement.clientWidth / window.innerWidth;
```
xripoll-at-wiris, line 485: Does this not work with browsers other than Safari? What's the behavior there?

packages/mathtype-html-integration-devkit/src/contentmanager.js
```diff
@@ -16,14 +16,15 @@ export default class ContentManager {
    * - The communication between the {@link ModalDialog} class and itself, if the content
    *   has been changed (hasChanges} method).
    * @constructs
-   * @param {Object} contentManagerAttributes - Object containing all attributes needed to
+   * @param {object} contentManagerAttributes - Object containing all attributes needed to
```
xripoll-at-wiris, line 19: There is a lot of linting changes; maybe those shouldn't be part of the PR since there's so many. In any case we can have another task/PR linting the devkit source files.

packages/mathtype-html-integration-devkit/src/contentmanager.js
icaparros-at-wiris: I checked that using an iPhone, and it is true. I removed the isSafari condition.

carla-at-wiris (approved):

carla-at-wiris (approved):

PR #625, opened 2022-07-21
[Monorepo] Migrate demos

carla-at-wiris:
## Description

This PR consist on migrating the html-integrations demo to the new framework under Nx. There were 3 main tasks regarding this change:
1. Add Nx to the current demos and the demos to the current Workspace.
2. Make the demos independent between them, to guarantee that the structure of the monorepo is not too mangled and each project within it is independent of the monorepo as a whole.
3. Create scripts to run demos with the packages locally and remotely.

### Also in this PR

* Nx and yarn help us to get rid of the `duplicated modules` CKEditor5 current issue.
* We created a private resources packages, which shares common code among demos. We want to maintain such package because the demos share, each one, around 300 code lines, and keeping those lines in a shared code will make us work agile, with less margin of error and faster.
* Updated the CHANGES.md file.
* Updated all the demos version to 1.1.0.
* Made all demos package.json `private` to avoid publishing them in the future.

## Steps to reproduce

You should try to run all the demos with the local and remote packages.

1. Run `yarn` on the project root.
2. To run demos with remote packages, run `nx start-remote <DEMO>` where demo can be
     * html-ckeditor4
     * html-ckeditor5
     * html-froala
     * html-generic
     * html-tinymce4
     * html-tinymce5
     * html-tinymce6
     * react-ckeditor5
     * react-froala
     * react-generic
     * react-tinymce5
     * angular-ckeditor5
     * angular-froala
     * angular-generic
     * angular-tinymce5
     
    > Note that, Tinymce6 wont work with the remote package, since is not published yet.
    > Froala demos wont work with the remote package either, since we are changing their name for the next release.
    > CKEditor5 for react and angular wont work with the local package for the editor integration limitations.
3. To run the demos with the local packages, run the following sequence of commands:
     ```sh
     nx build <PACKAGE>
     nx start <Demo>
     ```
     Where package can be:
     * ckeditor4
     * ckeditor5
     * froala
     * generic
     * tinymce4
     * tinymce5
     * tinymce6
    
     > Remember, that for each change you make in the package code, you'll have to run the build command again.

**All the commands can be executed from the project root. That way, you wont need to be changing folders.**

---

[#taskid 25399](https://wiris.kanbanize.com/ctrl_board/2/cards/25399/details/)


icaparros-at-wiris:
It is necesary to make a yarn link before run a demo

demos/html5/froala/package.json
```diff
@@ -1,16 +1,19 @@
...
-  "version": "1.0.4",
+  "version": "1.1.0",
+  "private": true,
   "description": "A simple html5 App integrating WIRIS MathType in a Froala4 rich text editor.",
   "main": "app.js",
   "scripts": {
-    "compile-package": "cd ../../../packages/mathtype-froala3/ && npm run compile -- npm --dev",
+    "prestart-remote": "yarn && yarn unlink @wiris/mathtype-froala && yarn install --force",
+    "start-remote": "webpack-dev-server --open",
+    "prestart": "yarn && yarn link @wiris/mathtype-froala",
```
icaparros-at-wiris, line 10: I am unable to open the froala demo. 

Without removing the 3 from `@wiris/mathtype-froala` I am able to open the demo correctly.

xripoll-at-wiris:
Local demos work fine.

The following behavior seems incorrect:

1. `yarn`
2. `nx build ckeditor4`
3. `nx start html-ckeditor4`
4. `nx start-remote html-ckeditor4`

After step 4, the demo launches, but if I check the `node_modules` in the root of the project, the `@wiris/mathtype-ckeditor4` package still points to the local folder.

carla-at-wiris:
The `yarn` command makes the wiris packages point to the local ones in the `node_modules` by default. 

It is in the demos behavior to decide whether or not to take that local dependency.

If you make changes inside one of the packages and then try the commands you mention, you'll see that the `start` command will take that change, while the `start-remote` command won't.

> Local demos work fine.
> 
> The following behavior seems incorrect:
> 
> 1. `yarn`
> 2. `nx build ckeditor4`
> 3. `nx start html-ckeditor4`
> 4. `nx start-remote html-ckeditor4`
> 
> After step 4, the demo launches, but if I check the `node_modules` in the root of the project, the `@wiris/mathtype-ckeditor4` package still points to the local folder.

xripoll-at-wiris:
Actually, I'm having further issues.

While trying to launch a local demo:

```
yarn
nx build ckeditor4
nx start html-ckeditor4
```

I looked into the `node_modules/@wiris` inside the demo folder and I see this:

`mathtype-ckeditor4 -> ../../../../../../../.config/yarn/link/@wiris/mathtype-ckeditor4`

And it does not contain the local changes.

Is that the same behavior you are getting? Maybe we have different global yarn configurations?

xripoll-at-wiris:
That issue has been resolved: turns out `yarn` was symlinking another instance of this repository that I had in another folder. This is a dangerous quirk of yarn workspaces, IMO.

Other issues:

- The devkit package is not symlinked when starting a demo.
- Maybe it would be worth it to use nx's task dependency system to compile the plugins before starting the demos.

xripoll-at-wiris (approved):

PR #626, opened 2022-07-27
[Monorepo] Add Cypress and Eslint Nx plugins

icaparros-at-wiris:
## Description

This PR adds Cypress and Eslint Nx plugins to the monorepo project. This plugins allows us to pass Tests and Linting on all workspace packages or an specific one using the nx commands.

## Steps to reproduce

To run tests from specific package:
1. Run `yarn` on the project root.
2. Run `nx build <package-name>` to build all packages.
3. Run `nx start <demo-name>` to start all demos.
4. Run `nx test <package-name>` to run all packages tests.

To lint specific package:
1. Run `yarn` on the project root.
2. Run `nx lint <package-name>` to run eslint on the package.

To lint all packages in parallel:
1. Run `yarn` on the project root.
2. Run `nx run-many --target=lint --all --parallel` to run eslint on all packages.

---

[#taskid 25401](https://wiris.kanbanize.com/ctrl_board/2/cards/25401/details/)

PR #627, opened 2022-07-28
[KB-25400] Adapt the workflows

xripoll-at-wiris:
## Description

This PR adapts the GitHub workflows of html-integrations to use Nx instead of Lerna.

In particular, it ports the `publish.yml` workflow (in charge of automatically publishing new versions of the npm packages) and the `jsdoc.yml` workflow (in charge of generating the static documentation site stored in devkit).
The third workflow, in charge of running the cypress tests, requires #626 to be merged.

## Steps to reproduce

To test these changes, you can do different things.

For the `jsdoc.yml` workflow, just make sure that the latest run is passing.

The `publish.yml` workflow is run on demand.
Unfortunately, there is no good way to test it as a workflow since it is intended to make permanent changes to the npm registry, which cannot be undone.
Thus, a possible alternative is to test the script it runs locally.
This requires some careful setup in order to avoid publishing packages accidentally to npm.

1. Make sure you are logged out by doing `npm logout` and `yarn logout`.
  This will prevent you from accidentally releasing any packages on npm.

2. Create the following files in a temporal folder outside of `html-integrations`:

  - `config.yaml`:
    ```yaml
    storage: ./storage
    auth:
      htpasswd:
        file: ./htpasswd
    packages:
      "@*/*":
        access: $all
        publish: $all
      "**":
        access: $all
        publish: $all
    logs:
      - { type: stdout, format: pretty, level: http }
    ```

  - `htpasswd`:
    ```
    # username:password pair.
    # Generated with `htpasswd -c htpasswd username`
    username:$apr1$gwQzvrYX$hLkvErWORju3C1RS2.QrW0
    ```

3. Make an `.npmrc` in your home folder with the following contents:

  ```
  # Publish to private registry
  registry=http://localhost:4873/

  # It is necessary to authenticate to publish in Verdaccio
  # Token for username="username", password="password"
  //localhost:4873/:_authToken="kSwBbr9NJVcgqZ4Kl4zufXavhs0H8aVgveIW8R9SftM="
  ```

4. Lift a Verdaccio instance from that same folder, and leave it running.
    ```sh
    npx verdaccio --config config.yaml
    ```

5. Create a branch off this one and work from that branch.

6. Run the script, providing the version combinations that you deem appropriate to test it out, for instance:

    ```sh
    ./scripts/publish.mjs devkit=1.9.0 ckeditor4=7.32.0
    ```

If done right, this should recreate the behavior described at the top of the `scripts/publish.mjs` file.
You can check that the publishing works by inspecting the `storage` folder within your temporary folder.
Check that this is indeed the case.

After testing, you should do some cleanup.

- The script will automatically create a new commit, some new tags (one per version), and push it all to origin.
You should remove the tags from origin (you can do that manually through the GitHub website) and from your local copy (`git tag -d <tagname>`), as well as the created Publish commit.

- You should delete the `.npmrc` file in your home folder.

- You can safely remove the temporal folder you have created earlier.

---

[#taskid 25400](https://wiris.kanbanize.com/ctrl_board/2/cards/25400/details/)


icaparros-at-wiris (approved):

.github/workflows/jsdoc.yml
```diff
@@ -68,24 +68,24 @@ jobs:
...
       #   with:
-      #     doc-folder: 'packages/mathtype-html-integration-devkit/out/'
+      #     doc-folder: 'packages/devkit/out/'
 
       # 07. Create a report with any broken link and/or image found.
       - name: Generate the broken links report
         run: |
-          npm run build-jsdoc-validation |& tee packages/mathtype-html-integration-devkit/out/report.txt 
-          cat packages/mathtype-html-integration-devkit/out/report.txt
+          yarn run build-jsdoc-validation |& tee packages/devkit/out/report.txt
```
carla-at-wiris, line 76: Any reason to use `yarn` commands instead of `nx` ones

xripoll-at-wiris: There is no need of interacting with the other packages and we are not caching this (so far), so might as well directly use yarn IMO.

carla-at-wiris (approved):

PR #630, opened 2022-08-02
fix: change webpack demo host to localhost

icaparros-at-wiris:
## Description

Chage webpack demo host to localhost. Windows doesn't display the demos when using the IP 0.0.0.0. Using `localhost` will work on Linux and Windows.

In IPv4 the address 0.0.0.0 is a non-routable meta-address used to designate an invalid, unknown or non applicable target. However, for practical reason, many client software treat 0.0.0.0 as localhost. For this reason we have decided to use `localhost` directly.

## Steps to reproduce

1. Run a demo.
2. Check if the url use localhost

---

[#taskid 21894](https://wiris.kanbanize.com/ctrl_board/2/cards/21894/details/)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 Last release of this project is was 30th of September 2021.
 
+### [Unrelease]
+
+  - fix: change webpack demo host to localhost
+
 ## 7.31.1 2022-08-18
 
   - docs: add deprecation warning message on froala and tinymce4 package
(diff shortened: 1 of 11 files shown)


xripoll-at-wiris (approved):

carla-at-wiris:
I think that react demos open with localhost by default, but angular demos are also defined to be opened with `0.0.0.0` instead of `localhost`. I believe that they should be changed too. That behavior is defined on the `package.json` file.

icaparros-at-wiris:
> I think that react demos open with localhost by default, but angular demos are also defined to be opened with `0.0.0.0` instead of `localhost`. I believe that they should be changed too. That behavior is defined on the `package.json` file.

Updated!

carla-at-wiris (approved):

PR #634, opened 2022-08-08
Added .gitattributes and .editorconfig files to force EOLs as LF

jguillen-at-wiris:
## Description

I added .gitattributes and .editorconfig files to force EOLs as LF.
These files will handle future pushes to get rid of CRLF EOLs in our documents.
Although, I have also normalized the files that still held CRLF EOLs, so no future cases should be raised, even with Windows users.

Further info:
https://www.aleksandrhovhannisyan.com/blog/crlf-vs-lf-normalizing-line-endings-in-git/

## Steps to reproduce

- After merging with master, re-cloning the repository in a new folder would just leave it ready.
- Just to make sure, if this is pulled or fast-forwarded into an already working tree, trigger the command:
`git add --renormalize .` 

---

[#taskid 27362](https://wiris.kanbanize.com/ctrl_board/2/cards/27362/details/)


carla-at-wiris (commented):
I see that you created multiple files to define the line endings, this means that it is not possible to create just one file for the hole project?

I also see that many webpack and readme files have changes as well but i fail to recognize the changes, could you tell me what is different?

jguillen-at-wiris:
The files need to be in the project root.
I set the files to the root of our monorepo and to the root of each of our plugin packages, so not only us working from our monorepo but everyone that might pull any of our packages benefits from this configuration. 

The changes in the readme files happened after normalizing the project. All these changed files used CRLF and now they use LF, so that's why you can't make the difference by sight.

.editorconfig
```diff
@@ -0,0 +1,4 @@
+root = true
+
+[*]
+end_of_line = lf
```
xripoll-at-wiris, line 4: Do we really want to force Windows users to use LF?

On IDEs (e.g. VS Code) it's probably not a problem, but on any basic plain text editor it might become a nuisance. In theory, using the `.gitattributes` should be enough so that no file is committed with CRLF.

See the note on [the EditorConfig wiki](https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#end_of_line):

> NOTE: if you want to use native line endings between different operating systems it is better not to set this option and leave that task to the VCS! In the future we might add a value like native for this scenario (cf [#226](https://github.com/editorconfig/editorconfig/issues/226)).

xripoll-at-wiris:
> I set the files to the root of our monorepo and to the root of each of our plugin packages, so not only us working from our monorepo but everyone that might pull any of our packages benefits from this configuration.

I don't think we ever established a principle for this, but we were going in the direction of assuming that developers would download the whole monorepo. I think having that many copies of the same config file is a bit redundant.

.editorconfig
jguillen-at-wiris: Indeed, to preserve and force our repository to use LF only, .gitattributes is all what we need.
The intention here is to preserve Windows developers from having their linters complaining on every line for that extra CR.
This should only affect editors that integrate the .editorconfig file settings in their configuration, leaving the rest to perform natively.

xripoll-at-wiris (approved):

carla-at-wiris (approved):

jgonzalez-at-wiris:
The structure of the project changed so the suggested changes aren't valid anymore. Closing this issue, we'll retake it in the future.

PR #674, opened 2022-10-25
[KB-28734] Workflow for the staging of html-integrations demos

xripoll-at-wiris:
## Description

This PR does the following:

- Adapts the html-integrations demos so they can be built and served as static files and not only from the Webpack/Angular/React development servers.
- Adds a workflow that runs on every push and deploys a staging environment for the pushed branch.
- Adds a workflow that removes the deployed staging environment for a given branch when such branch is deleted.

## Steps to reproduce

It is important to test:

- That the staging demos deployed by this branch (with format https://integrations.wiris.kitchen/KB-28734/[FRAMEWORK]/[EDITOR]/) are deployed correctly and work as expected.
- That the development server demos (i.e. the demos being served by Webpack/Angular/React up until now) still work correctly in the local environment, and the instructions to do so match.

Try at least some of the demos on your own machine, to see if they still work properly.

**Note**: the CKEditor 5 demo for HTML does currently not work. This should be fixed in a future task. 

---

[#taskid KB-28734](https://wiris.kanbanize.com/ctrl_board/2/cards/KB-28734/details/)


icaparros-at-wiris:
I created a new branch from `KB-28734` and the staging demos have been deployed correctly. 

When I delete the branch directly from the GitHub page, the `Dismantle staging` the workflow has not been executed.

I made something wrong in order to dismantle the staging demos?

xripoll-at-wiris:
> When I delete the branch directly from the GitHub page, the Dismantle staging the workflow has not been executed.

Yes, that is currently expected; I forgot to mention that.

The delete workflow uses the `delete` event which only gets triggered if it is present in the main branch of the project. So, this will not work until we merge this PR, unfortunately.

If we deem it necessary, I can split this PR in two parts and merge the delete workflow before this one.

icaparros-at-wiris:
> If we deem it necessary, I can split this PR in two parts and merge the delete workflow before this one.

It's ok for me with the explanation Just want to be sure that is not an issue.

We will have to delete the `test-staging` demos manually this time.

icaparros-at-wiris (approved):

xripoll-at-wiris:
> I'm just missing some documentation or steps to follow for future references, or for our QA members.

Good idea! I've just added some. Let me know if you feel like something is missing.

carla-at-wiris (approved):

icaparros-at-wiris (approved):

PR #687, opened 2022-11-09
fix(ckeditor5): chemtype formulas Drag & Drop

icaparros-at-wiris:
## Description

Chem equations disappear when you drag them into another position inside the CKEditor. This only happens with the 8.0.0 version of our plugin.

Due an incorrect DOMPurify implamentation on CKEditor5 the ChemType formulas were detected as XSS attacks when dropped.

## Steps to reproduce

1. Open CKEditor5 demo.
2. Insert a ChemType formula and text.
3. Drag the ChemType formula to another position.
4. 

---

[#taskid 30339](https://wiris.kanbanize.com/ctrl_board/2/cards/30339/details/)

### DIFF
--- a/packages/ckeditor5/src/plugin.js
+++ b/packages/ckeditor5/src/plugin.js
@@ -254,12 +254,7 @@ export default class MathType extends Plugin {
       let formula = processor.toData(viewDocumentFragment) || '';
 
       // And obtain the complete formula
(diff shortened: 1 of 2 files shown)


carla-at-wiris:
The demo appears without the initial content and the cursor is not focusing the demo, so the user has to click on the editor on purpose

carla-at-wiris (approved):

PR #754, opened 2023-01-12
KB-31663: TinyMCE 5 & 6 demos share module and confuses Nx

icaparros-at-wiris:
## Description

Currently, Tinymce5 & 6 demos copy the Tinymce module to the dist folder in order to adapt demos for static deployment. Both plugins use the same module folder name (tinymce in the dist folder). NX detects as the folders as the same workspace in two different places and throws an error message.

Changing the module folder names on each Tinymce fix the issue.

## Steps to reproduce

1. Install dependencies with yarn
2. Execute:
```
nx build tinymce5
nx start html-tinymce5
nx build tinymce6
nx start html-tinymce6
```
3. Close both demos
4. Try to restart any demo.
---

[#taskid 31663](https://wiris.kanbanize.com/ctrl_board/2/cards/31663/details/)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,7 @@ Last release of this project is was 30th of September 2021.
 ### 8.1.1 2023-01-11
 
   - Fix: Generic demos on staging are now in english. #KB-31349
+  - Fix: tinymce folder names confuse nx. #KB-31663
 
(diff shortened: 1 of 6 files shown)


carla-at-wiris (approved):
Very clear solution!

PR #758, opened 2023-01-17
Fix language not being read from server editor parameters

xripoll-at-wiris:
## Description

Fix a bug that stopped the front end from using the language set in configuration.ini.

## Steps to reproduce

Open a demo from the plugins project and modify the configuration.ini to use a language different than the default one. The language of the modal and the editor should be that defined in the configuration.ini file. Detailed instructions can be found on the associated card.

---

[#taskid 30880](https://wiris.kanbanize.com/ctrl_board/2/cards/30880/details/)


carla-at-wiris (approved):
The code looks fine, and the demo works properly!

PR #759, opened 2023-01-17
Make control buttons accessible for Keyboard

carla-at-wiris:
## Description

Minimize, maximize, stack and close buttons on the top right of the modal editor were not accessible through the Keyboard, and therefore made the MathType not user-friendly. This PR fixes the issue by changing the logic of the Keyboard interaction with these buttons.

## Steps to reproduce

1. Install dependencies with `yarn`.
2. Build the desired package with `nx build <PACKAGE>`.
3. Start the desired demo with `nx start html-<PACKAGE>`.
4. Verify that the top right buttons are accessible through the keyboard. Verify that they work. Verify that once the modal is minimized or maximized, the buttons are also working

> Since the issue is reproducible in all packages and demos, it'll be preferible to perform the tests in more than one demo & package.

---

[#taskid 30749](https://wiris.kanbanize.com/ctrl_board/2/cards/30749/details/)

### DIFF
--- a/packages/devkit/src/contentmanager.js
+++ b/packages/devkit/src/contentmanager.js
@@ -570,22 +570,35 @@ export default class ContentManager {
           this.editor.focus();
           keyboardEvent.stopPropagation();
           keyboardEvent.preventDefault();
+        } else if (document.querySelector('[title="Manual"]') === document.activeElement) {
+          // Focus is on minimize button (_).
+          this.modalDialogInstance.closeDiv.focus();
+          keyboardEvent.stopPropagation();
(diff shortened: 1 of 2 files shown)


xripoll-at-wiris:
By tabbing, the order seems to be

Close button -> Full-screen button -> Minimize button

which is right to left. Wouldn't it make more sense to have it be left to right?

Other than that, the feature seems to work perfectly.

xripoll-at-wiris (approved):

icaparros-at-wiris (approved):

PR #760, opened 2023-01-18
Show a friendly message when going offline with MathType

carla-at-wiris:
## Description

MathType is an online tool, but it gives an error with no explanation when going offline. This PR adds a modal window that warns the user that MathType can't work offline.

## Steps to reproduce

1. install dependencies with `yarn`.
2. Build the plugin with `nx build <PACKAGE>`.
3. Start the desired demo with `nx start <FRAMEWORK>-<PACKAGE>`.
4. Turn off the internet connection.
5. Click the MT/CT button or double-click a formula.
6. A modal window should appear, warning the user that MathType can not work offline.
7. Validate that the modal that appears also matches [the one created by the Product Design team](https://www.figma.com/file/L0fMclgbWUTPaWBQDKIdzX/Pop-up-integrations?node-id=0-1&t=GGG5PNzjXMxzuMQA-0).

---

[#taskid 12739](https://wiris.kanbanize.com/ctrl_board/2/cards/12739/details/)


packages/devkit/src/serviceprovider.js
```diff
@@ -173,28 +173,30 @@ export default class ServiceProvider {
...
+          httpRequest.send(null);
+        }
 
-      return httpRequest.responseText;
-    }
-    return '';
+        return httpRequest.responseText;
+      }
+      return '';
+    } else window.alert("MathType can not work offline");
```
xripoll-at-wiris, line 199: "can not" should be "cannot"

xripoll-at-wiris (changes requested):
I'm not sure this is a good user experience. The message only works when the connection is lost before loading the page, and an alert seems to be intrusive.

I think, ideally, we should still open the modal, but display an error message inside.

@jguillen-at-wiris maybe we should get Product involved in this.

carla-at-wiris:
> I'm not sure this is a good user experience. The message only works when the connection is lost before loading the page, and an alert seems to be intrusive.
> 
> I think, ideally, we should still open the modal, but display an error message inside.
> 
> @jguillen-at-wiris maybe we should get Product involved in this.

I agree, unfortunately, the modal can not be opened properly without internet connection. Maybe we can introduce the message in some way before the errors appear, even though the modal will not have all the MathType features. I also think that product should be involved in this @jguillen-at-wiris

packages/devkit/src/integrationmodel.js
```diff
@@ -285,17 +311,29 @@ export default class IntegrationModel {
...
    * actions from integration part before the formula is edited.
    */
   openExistingFormulaEditor() {
-    this.core.editionProperties.isNewElement = false;
-    this.core.openModalDialog(this.target, this.isIframe);
+    if (window.navigator.onLine) {
+      this.core.editionProperties.isNewElement = false;
+      this.core.openModalDialog(this.target, this.isIframe);
+    } else {
+      // alert('PARARA');
```
xripoll-at-wiris, line 333: PARARA

xripoll-at-wiris (commented):
Good job! It works very well. The only quirk I found is that on Froala, the toolbar underneath the new modal stays visible and can be clicked:

html-froala

xripoll-at-wiris (approved):

xripoll-at-wiris:
Remember to rebase! You have conflicting changes with `master` now that I notice.

icaparros-at-wiris (changes requested):
Great job!

On Froala the alert svg and the modal close/minimize/maximize icons are not been displayed.
Do you know If this is a bug fixed in another card?

image
image

icaparros-at-wiris (approved):

PR #784, opened 2023-02-21
[KB-32689] Add new initial formulas to demos

xripoll-at-wiris:
## Description

This PR updates the initial content of the editors in the demos to display four different formulas.

## Steps to reproduce

Start each of the demos and check that the content is correct. You can check the intended content in the [linked card](https://wiris.kanbanize.com/ctrl_board/2/cards/32689/details/).

Pay special attention to whether the formulas are using MathType/ChemType and Keyboard/Hand when opened.

---

[#taskid 32689](https://wiris.kanbanize.com/ctrl_board/2/cards/32689/details/)


carla-at-wiris (commented):
Since each formula is the same, couldn't we store them in a common resource area, and then import the content on the HTML? That way we wouldn't need to have duplicated code in all the demos.

carla-at-wiris (approved):

PR #795, opened 2023-03-20
[KB 33529] - Editor becomes unusable when reducing the screen to 550px width

icaparros-at-wiris:
## Description

When users open the MathType editor with a desktop window smaller than 550 px width, the MathType detects as user is on mobile. This PR fix this issue detecting correctly If user is on Desktop.

## Steps to reproduce
 
1. Open a demo from `html-integrations`.
2. Reduce the screen to the minimum.
3. Press the MathType icon to insert an equation.

The MathType editor will be open correctly.

---

[#taskid 33529](https://wiris.kanbanize.com/ctrl_board/2/cards/33529/details/)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,10 @@ Last release of this project is was 30th of September 2021.
 
 ### Unreleased
 
+  - Fix: detect if device is mobile. #KB-33529
+
+### 8.2.6 2023-03-17
+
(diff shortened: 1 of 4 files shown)


carla-at-wiris (commented):
I'm missing the update of the changes.md file. 
Apart from that, everything looks fine!

PR #800, opened 2023-04-18
Moodle not sending the correct course and category to Telemetry

carla-at-wiris:
## Description

The Telemetry events were not sending the correct Moodle course and category. Those were sent by the back-end, and were set just once, when the filter is loaded, which caused the values to never change from their original value.

This PR changes the way the data is sent from the back-end to the front-end + moodle in both TinyMCE and Atto plugins.

## Steps to reproduce

1. Go to any staging demo.
2. Validate that it's sending the proper data to Tableau.
3. Open a Moodle instance and validate that it's also sending the correct data to Tableau.

---

[#taskid 34608](https://wiris.kanbanize.com/ctrl_board/2/cards/34608/details/)

### DIFF
--- a/packages/devkit/src/integrationmodel.js
+++ b/packages/devkit/src/integrationmodel.js
@@ -274,10 +274,10 @@ export default class IntegrationModel {
       lms = {
         nam: 'moodle',
         fam: 'lms',
-        ver: Configuration.get('versionMoodle'),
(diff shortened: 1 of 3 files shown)


xripoll-at-wiris (approved):
Remember to rebase from `master`!

PR #805, opened 2023-04-26
fix: convert special symbols to entities when using ck5 getData()

icaparros-at-wiris:
## Description

When user set a custom toolbar with the buttons `<` or `>`, for example:

```
"<toolbar ref='general'><tab ref='symbols' empty='true'><section rows='3'><item ref='&gt;' extra='false'/></section></tab></toolbar>"
```
:warning: We use `&gt;` and `&lt;` instead  `<` and `>` to avoid conflicts with `showimage` service. 

The custom toolbar xml is added on the `<annotation>` tag from inserted formulas.

CKEditor5 convert the `&gt;` and `&lt;` to `<` and `>` again, causing issues with `showimage` service on users that have custom tags with this two buttons set when they use the getData() function.

This issue have been fixed on this PR and now the special characters are converted to HTML Entity (decimal) when using getData() function.

## Steps to reproduce

1. Set next custom toolbar on editorParameters on CKEditor5: 
```
"<toolbar ref='general'><tab ref='symbols' empty='true'><section rows='3'><item ref='&gt;' extra='false'/></section></tab></toolbar>"
```
3. Open the CKEditor5 demo
4. Insert a formula
5. Open the inspector
6. Insert next code on the inspector console:
```
const data = editor.getData();
editor.setData(data);
```

No error appears on inspector console.

---

[#taskid 34675](https://wiris.kanbanize.com/ctrl_board/2/cards/34675/details/)

### DIFF
--- a/demos/html/ckeditor5/src/app.js
+++ b/demos/html/ckeditor5/src/app.js
@@ -34,11 +34,12 @@ ClassicEditor
     plugins: [Essentials, Paragraph, Bold, Italic, MathType, Alignment],
     toolbar: ['bold', 'italic', 'MathType', 'ChemType', 'alignment:left', 'alignment:center', 'alignment:right'],
     // language: 'de',
-    // mathTypeParameters: {
-    //   editorParameters: { language: 'es' }, // MathType config, including language
-    // },
+    mathTypeParameters: {
+      editorParameters: { toolbar: "<toolbar ref='general'><tab ref='symbols' empty='true'><section rows='3'><item ref='&gt;' extra='false'/><item ref='&lt;' extra='false'/></section></tab></toolbar>" }, // MathType config, including language
+    },
   })
   .then((editor) => {
+    alert('Remove last commit on merge');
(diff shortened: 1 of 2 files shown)


xripoll-at-wiris:
When adding a custom toolbar, I cannot manage to open the MathType editor on the CKEditor 4 for HTML demo. The editor stays blank and I get an `Uncaught Expected attribute name` error in console.

icaparros-at-wiris:
> When adding a custom toolbar, I cannot manage to open the MathType editor on the CKEditor 4 for HTML demo. The editor stays blank and I get an `Uncaught Expected attribute name` error in console.

The toolbar xml set on the steps to reproduce was incorrect.

I updated the example with this one:

```
"<toolbar ref='general'><tab ref='symbols' empty='true'><section rows='3'><item ref='&gt;' extra='false'/></section></tab></toolbar>"
```

carla-at-wiris (approved):

PR #808, opened 2023-05-08
Develop the main set of features of the viewer

xripoll-at-wiris:
## Description

This PR adds the main set of features of the new viewer.

To check that everything is correct, compare to the specification (see [card](https://wiris.kanbanize.com/ctrl_board/2/cards/34546/details/)) and test the demo included in the package itself (`yarn serve` in `packages/viewer`).

Please don't hesitate to ask or comment on any possible improvement or missing feature.

---

[#taskid 34546](https://wiris.kanbanize.com/ctrl_board/2/cards/34546/details/)

### DIFF
--- a/packages/viewer/src/properties.ts
+++ b/packages/viewer/src/properties.ts
@@ -1,0 +1,256 @@
+import { configurationJson, StatusError } from './services';
+
+// Helper types for Config below
+type Viewer = 'image' | 'none';
+type Wirispluginperformance = 'true' | 'false';
+
+/**
+ * Type representing all the configuration for the viewer.
+ */
+export type Config = {
+  editorServicesRoot?: string,
+  editorServicesExtension?: string,
+  backendConfig?: {
+    wirispluginperformance?: Wirispluginperformance,
+    wiriseditormathmlattribute?: string,
+  },
+  dpi?: number,
+  element?: string,
+  lang?: string,
+  viewer?: Viewer,
+  zoom?: number,
+};
+
+/**
+ * Fallback values for the configurations that are not set.
+ */
+const defaultValues: Config = {
+  editorServicesRoot: 'https://www.wiris.net/demo/plugins/app/',
+  editorServicesExtension: '',
+  backendConfig: {
+    wirispluginperformance: 'true',
+    wiriseditormathmlattribute: 'data-mathml'
+  },
+  dpi: 96,
+  element: 'body',
+  lang: 'en',
+  viewer: 'none',
+  zoom: 1,
+}
+
+/**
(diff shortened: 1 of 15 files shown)


carla-at-wiris (commented):
I see that there are some TODOs over the project, are they going to stay this way and be finished after the first release or should they be fixed before it?

I'm failing to see the md5, is it not going to be implemented?

carla-at-wiris (commented):
I see that there are some TODOs over the project, are they going to stay this way and be finished after the first release or should they be fixed before it?

I'm failing to see the md5, is it not going to be implemented?

xripoll-at-wiris:
Hi @carla-at-wiris!

Indeed, the MD5 feature was missing. I implemented it now, so check again if you can.

Regarding the TODOs, I think a couple of them warrant creating new cards, and some others don't have an immediate impact nor solution.

carla-at-wiris (commented):
Is the index.html planned to be published? Or is it only for development purposes?

carla-at-wiris (approved):

packages/viewer/src/services.ts
```diff
@@ -0,0 +1,178 @@
...
+/**
+ * Calls the endpoint servicename and returns its response.
+ * @param {object} query - Object of parameters to pass as the body request or search parameters.
+ * @param {string} serviceName - Name of the service to be called.
+ * @param {string} serverURL - Url of the server where we want to call the service.
+ * @returns {Promise<Response>} The request response.
+ */
+export async function callService(query: object, serviceName: string, method: MethodType, serverURL: string) : Promise<any> {
+  try {
+    const url = new URL(serviceName, serverURL);
```
icaparros-at-wiris, line 50: PHP services need to add the .php extension at the end of the service name. 

If we want to allow users to change the backend could be necessary to add a new property to set the backend technology and add the corresponding suffix on each service call.

xripoll-at-wiris: Good catch! Fixed

packages/viewer/src/properties.ts
```diff
@@ -0,0 +1,242 @@
...
+ */
+export abstract class Properties {
+
+  static render: () => Promise<void> = async () => {};
+
+  // Flag for the static parameters that access the backend.
+  private static backendObtained: boolean = false;
+
+  private static waitForBackend() {
+    while (!this.backendObtained);
```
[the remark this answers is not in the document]
xripoll-at-wiris: The try that wraps the request has a finally clause which sets `backendObtained` to true, so in theory this loop should always finish one way or another.

icaparros-at-wiris (approved):

PR #815, opened 2023-05-24
Fix strange behavior when the caret at the beginnig

carla-at-wiris:
## Description

When the caret was at the beginning of the editor where there's a formula, instead of opening a new formula, it edits the one on top. This PR fixes the mentioned behavior by changing the nodes where the formulas should be added.

This bug was happening also in between 2 formulas when they were at the beginning of the editor.

## Steps to reproduce

1. Install dependencies with `yarn`.
2. Build the desired demo with `nx build <PACKAGE>`.
3. Start the demo with `nx start html-<PACKAGE>`.
4. Empty the content of the text area.
5. Insert a formula.
6. Place the caret at the beginning of the text are and insert a formula.
7. Verify that when clicking on the MT/CT button, the first formula was not being edited. Verify also that the new added formula is placed where it should.

> The issue was present on all editors but CKEditor5. They should all be tested.

---
[#taskid 21754](https://wiris.kanbanize.com/ctrl_board/2/cards/21754/details/)

### DIFF
--- a/packages/devkit/src/util.js
+++ b/packages/devkit/src/util.js
@@ -791,6 +791,24 @@ export default class Util {
         const position = range.startOffset;
 
         if (node.childNodes[position]) {
+
+          // In case that a formula is detected but not selected,
+          // we create an empty span where we could insert the new formula.
+          if (range.startOffset === range.endOffset) {
+            if (position !== 0 && node.childNodes[position - 1].localName === 'span' && node.childNodes[position].classList.contains('Wirisformula')) {
+              node.childNodes[position - 1].remove();
+              return Util.getSelectedItem(target, isIframe, forceGetSelection);
+            }
+            else if (node.childNodes[position].classList.contains('Wirisformula')) {
+              if ((position > 0 && node.childNodes[position - 1].classList.contains('Wirisformula')) || position === 0 ) {
+                var emptySpan = document.createElement('span');
+                node.insertBefore(emptySpan, node.childNodes[position]);
+                return {
+                  node: node.childNodes[position],
+                }
+              }
+            } 
+          }
           return {
             node: node.childNodes[position],
           };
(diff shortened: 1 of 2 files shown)


packages/devkit/src/util.js
```diff
@@ -775,6 +775,8 @@ export default class Util {
       }
 
       const node = range.startContainer;
+      console.log('RANGE');
+      console.log(range);
```
xripoll-at-wiris, line 779: Remember to remove the logs!

packages/devkit/src/util.js
```diff
@@ -789,8 +791,33 @@ export default class Util {
 
       if (node.nodeType === 1) { // ELEMENT_NODE.
         const position = range.startOffset;
+        console.log('POSITION');
+        console.log(position);
+        console.log('NODE');
+        console.log(node);
+        console.log('WIRISPLUGIN')
+        console.log(WirisPlugin)
```
xripoll-at-wiris, line 799: Here too

packages/devkit/src/util.js
```diff
@@ -789,8 +791,33 @@ export default class Util {
...
+          // we create an empty span where we could insert the new formula.
+          if (range.startOffset === range.endOffset) {
+            if (position !== 0 && node.childNodes[position - 1].localName === 'span' && node.childNodes[position].classList.contains('Wirisformula')) {
+              node.childNodes[position - 1].remove();
+              return Util.getSelectedItem(target, isIframe, forceGetSelection);
+            }
+            else if (node.childNodes[position].classList.contains('Wirisformula')) {
+              if ((position > 0 && node.childNodes[position - 1].classList.contains('Wirisformula')) || position === 0 ) {
+                var emptySpan = document.createElement('span');
+                // const emptySpan = document.createTextNode('');
```
xripoll-at-wiris, line 813: This line can probably be removed as well.

xripoll-at-wiris (changes requested):

xripoll-at-wiris (approved):

PR #817, opened 2023-05-25
fix (ckeditor5): improve inline widget insertion

### LINKED ISSUE #811: [CKEditor 5] Improve inline widget insertion
## Description

While investigating the bug reported in https://github.com/ckeditor/ckeditor5/issues/14106, we noticed some issues with MathType integration. The issue described in the bug is fixed for inline widgets. To make it work for MathType, it needs to:

1. Change the schema registration from

    https://github.com/wiris/html-integrations/blob/29fd293e7a7dfaacd4daed0d26fe2b5efebc7c35/packages/ckeditor5/src/plugin.js#L201C17-L206

    to:

    ```js
    schema.register('mathml', {
      inheritAllFrom: '$inlineObject',
      allowAttributes: ['formula'],
    });
    ```
    
    For more details see https://ckeditor.com/docs/ckeditor5/latest/framework/tutorials/implementing-an-inline-widget.html#defining-the-schema.

2. Change the way you modify the model to insert the `<mathml>` element. Instead of using `writer.insert()` as shown below, use `model.insertObject()` instead.

    https://github.com/wiris/html-integrations/blob/29fd293e7a7dfaacd4daed0d26fe2b5efebc7c35/packages/ckeditor5/src/integration.js#L161

    https://github.com/wiris/html-integrations/blob/29fd293e7a7dfaacd4daed0d26fe2b5efebc7c35/packages/ckeditor5/src/integration.js#L183

    For more information see https://ckeditor.com/docs/ckeditor5/latest/framework/tutorials/implementing-an-inline-widget.html#command

## Environment

Latest

CKEditor 5

## Steps to reproduce

1. Use the latest `@ckeditor/ckeditor5-engine` package from the CKEditor 5 repository (because the fix for the bug linked above has not yet been released).
2. Follow the reproduction steps in https://github.com/ckeditor/ckeditor5/issues/14106, but instead of using inline image, use mathml.

## Expected result

Text formatting is preserved after inserting the widget.

icaparros-at-wiris:
## Description

CKEditor5 Team detected the bug https://github.com/ckeditor/ckeditor5/issues/14106:

> Currently, when non-defualt font properties are selected, and then a widget (e.g. image, math formula) is inserted into the document, the font reverts back to its default state.
> 
> This behavior could be improved in such a way, that the selected font family, size, background color etc. would be preserved upon inserting a widget.

This PR adds CKEditor Team proposed changes to work better with the recent improvements on CKEditor5.

> The CKEditor5 is fixed on the latest release of `@ckeditor/ckeditor5-engine` but this improvement will work correctly on oldest versions.

## Steps to reproduce

1. Use the latest `@ckeditor/ckeditor5-engine` package from the CKEditor 5 repository (because the fix for the bug linked above has not yet been released).
2. Follow the reproduction steps in https://github.com/ckeditor/ckeditor5/issues/14106, but instead of using inline image, use mathml.

---

Closes #811

[#taskid 35934](https://wiris.kanbanize.com/ctrl_board/2/cards/35934/details/)

### DIFF
--- a/packages/ckeditor5/src/integration.js
+++ b/packages/ckeditor5/src/integration.js
@@ -2,7 +2,6 @@ import IntegrationModel from '@wiris/mathtype-html-integration-devkit/src/integr
 import Util from '@wiris/mathtype-html-integration-devkit/src/util';
 import Configuration from '@wiris/mathtype-html-integration-devkit/src/configuration';
 import Latex from '@wiris/mathtype-html-integration-devkit/src/latex';
-import Parser from '@wiris/mathtype-html-integration-devkit/src/parser';
 import MathML from '@wiris/mathtype-html-integration-devkit/src/mathml';
 import Telemeter from '@wiris/mathtype-html-integration-devkit/src/telemeter';
 
@@ -158,7 +157,7 @@ export default class CKEditor5Integration extends IntegrationModel {
         const viewSelection = this.core.editionProperties.selection || this.editorObject.editing.view.document.selection;
         const modelPosition = this.editorObject.editing.mapper.toModelPosition(viewSelection.getLastPosition());
 
-        writer.insert(modelElementNew, modelPosition);
+        this.editorObject.model.insertObject(modelElementNew, modelPosition);
 
         // Remove selection
         if (!viewSelection.isCollapsed) {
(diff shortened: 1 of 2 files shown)


carla-at-wiris (approved):

Witoso:
When do you plan to release the changes? Asking from the CKEditor side on behalf of our clients.

PR #825, opened 2023-06-26
feat: add vue + ckeditor5 demo

icaparros-at-wiris:
## Description

This PR adds a new demo for MathType + CKEditor5 using Vue as a framework.

This new demo is made in TypeScript and use Vite to bundle all the code in order to use the recommended stack for Vue.

## Steps to reproduce

Use the next commands on the terminal:

```
yarn
nx start vue-ckeditor5
```

---

[#taskid KB36629](https://wiris.kanbanize.com/ctrl_board/2/cards/36629/details/)


carla-at-wiris (changes requested):
The staging HTML demos are not being generated and the workflow fails

carla-at-wiris (changes requested):
* The Vue staging demo is blank, and the following errors appear on the console:
image
* The `yarn.lock` should not be removed. At most, updated.

carla-at-wiris (approved):
Remember to try to not publish the yarn.lock file

xripoll-at-wiris:
The [HTML CKEditor 4 demo on staging](https://integrations.wiris.kitchen/KB-36629/html/ckeditor4/) has errors:
image

carla-at-wiris (changes requested):
All the `react` and `angular` staging demos are failing. This shouldn't happen except with the Froala + Angular demo.

demos/vue/ckeditor5/README.md
```diff
@@ -0,0 +1,153 @@
...
+
+## Recommended IDE Setup
+
+[VSCode](https://code.visualstudio.com/) + [Volar](https://marketplace.visualstudio.com/items?itemName=Vue.volar) (and disable Vetur) + [TypeScript Vue Plugin (Volar)](https://marketplace.visualstudio.com/items?itemName=Vue.vscode-typescript-vue-plugin).
+
+## How to run the demo
+
+```sh
+$ npm install
+$ npm run dev
```
carla-at-wiris, line 18: Isn't there a way to use `yarn` + `nx`, as we do in the rest of the demos?

packages/ckeditor5/src/plugin.js
```diff
@@ -24,8 +24,8 @@ import StringManager from '@wiris/mathtype-html-integration-devkit/src/stringman
 import { MathTypeCommand, ChemTypeCommand } from './commands';
 import CKEditor5Integration from './integration';
 
-import mathIcon from '../theme/icons/formula.svg';
-import chemIcon from '../theme/icons/chem.svg';
+import mathIcon from '../theme/icons/ckeditor5-formula.svg';
+import chemIcon from '../theme/icons/ckeditor5-chem.svg';
```
carla-at-wiris, line 28: I'm missing an explanation of why do we have specific icons for CKEditor5, instead of using the same ones we use in the other demos. This is for future references, in case we bump into this part of the code and are unaware of why is this happening.

carla-at-wiris (changes requested):

packages/ckeditor5/src/plugin.js
icaparros-at-wiris: For some reason the package `vite-plugin-ckeditor5` force us to add the prefix `ckeditor5-` on svg used on ckeditor5.

demos/vue/ckeditor5/README.md
icaparros-at-wiris: Updated :D

carla-at-wiris (approved):

PR #828, opened 2023-06-28
Alignment issues in TinyMCE 6 and Froala

xripoll-at-wiris:
## Description

In TinyMCE 6, formulas lose alignment when clicked.

In Froala, formulas added in the demo by default came without proper alignment.

This PR fixes both of these issues.

## Steps to reproduce

1. Open up the TInyMCE 6 and Froala demos.
2. Check that the formulas are properly aligned
3. Click on any formula, and check that it stays aligned.

---

[#taskid 34325](https://wiris.kanbanize.com/ctrl_board/2/cards/34325/details/)


carla-at-wiris (changes requested):
TinyMCE alignment is all right.
Froala is still facing the same issues as before.

carla-at-wiris (approved):

PR #830, opened 2023-07-04
Get froala api key from secrets on staging

icaparros-at-wiris:
## Description

This PR gets the froala api key from github secrets when deploying the [staging demos](https://integrations.wiris.kitchen/stable/html/froala/).

If we don't set the api key the froala editor will self-destruct after a while when not using the localhost.

---

[#taskid 36829](https://wiris.kanbanize.com/ctrl_board/2/cards/36829/details/)

### DIFF
--- a/.github/workflows/deploy-staging.yml
+++ b/.github/workflows/deploy-staging.yml
@@ -87,6 +87,12 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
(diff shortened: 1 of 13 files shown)


carla-at-wiris (changes requested):
I see that the angular and react demos still have the license desactivated

carla-at-wiris (approved):

PR #838, opened 2023-07-26
Add GitHub information to the demo

carla-at-wiris:
## Description

This PR adds the information of the GitHub branch and commit to the demo pages.

## Steps to reproduce

1. Go to all staging demos.
2. Validate that they all have the new information matching this current branch and its latest commit.

---

[#taskid 34114](https://wiris.kanbanize.com/ctrl_board/2/cards/34114/details/)

### DIFF
--- a/packages/res/git-data.mjs
+++ b/packages/res/git-data.mjs
@@ -1,0 +1,24 @@
+import { execSync } from 'child_process';
+import * as fs from 'fs';
+import path from 'path';
+import { fileURLToPath } from 'url';
+
+let x1 = 'unknown', x2 = 'unknown';
+
+// Retrieve GitHub branch and commit hash
+x1 = execSync('git rev-parse --abbrev-ref HEAD').toString().trim();
+x2 = execSync('git rev-parse --short HEAD').toString().trim();
+
+// Converting JS object to JSON string
+var jsonData = JSON.stringify({branch: x1, hash: x2});
+
+// Get current directory.
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+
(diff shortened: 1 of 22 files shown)


packages/res/git-data.mjs
```diff
@@ -0,0 +1,24 @@
...
+
+// Converting JS object to JSON string
+var jsonData = JSON.stringify({branch: x1, hash: x2});
+
+// Get current directory.
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+
+// Write data in json file.
+fs.writeFile(__dirname + '/git-data.json', jsonData, (err) => {
```
xripoll-at-wiris, line 20: If you've already imported `path`, might as well use `path.join` here instead of + for obtaining the full path.

xripoll-at-wiris (changes requested):

xripoll-at-wiris:
What do we think about adding the branch info to the `<title>` of the page, so we can tell apart which demo is which just by looking at the tab?

I think it would be nice but not if it hides the rest of the info (framework and editor).

carla-at-wiris:
> What do we think about adding the branch info to the `<title>` of the page, so we can tell apart which demo is which just by looking at the tab?
> 
> I think it would be nice but not if it hides the rest of the info (framework and editor).

The title of the page is already way too long with the information of the framework and editor, it almost never shows up all the title on the tab, so adding the branch on it would not give us any more information in a regular basis. 
To see the branch in the title, we would have to place the cursor on the tab and wait until the full title shows up, which is more or less the same as opening the page.

xripoll-at-wiris (approved):

PR #865, opened 2023-11-30
feature: Include mechanism to force hand mode when editor is opened

jgonzalez-at-wiris:
## Description

This PR includes a mechanism to force the editor to be opened in hand mode.

## Steps to reproduce

To activate this feature, you'll need to modify how the Wiris editor is initialized. You can do so modifying line 26 in `app.js`:

```
window.wrs_int_init(editableDiv, toolbarDiv, undefined, { forcedHandMode: true }); // will activate forced hand mode
```

You can try this other modifications to experiment the different behaviours:

```
window.wrs_int_init(editableDiv, toolbarDiv); // won't activate forced hand mode
window.wrs_int_init(editableDiv, toolbarDiv, undefined); // won't activate forced hand mode
window.wrs_int_init(editableDiv, toolbarDiv, undefined, {}); // won't activate forced hand mode
window.wrs_int_init(editableDiv, toolbarDiv, undefined, { forcedHandMode: false }); // won't activate forced hand mode
```
As you can see, the configuration object has been included as an extra argument to the `wrs_int_init` method. I decided to do so to avoid changing the content of the rest of the arguments and make sure we won't break any existing production code from our clients). I know it's ugly having an undefined argument, but since I'm not sure when it's really used or not, I've prefered to keep it that way for the sake of retro compatibility, as I mentioned before.

Anny commentary regarding variable names, and how the configuration argument is sent all the way to the ContentManager will be appreciated.
---

[#taskid 41415](https://wiris.kanbanize.com/ctrl_board/2/cards/41415/details/)

### DIFF
--- a/packages/devkit/src/integrationmodel.js
+++ b/packages/devkit/src/integrationmodel.js
@@ -170,6 +170,8 @@ export default class IntegrationModel {
         }
       });
     }
+
+    this.editorOptions = integrationModelProperties.editorOptions || {};
   }
 
   /**
@@ -204,6 +206,7 @@ export default class IntegrationModel {
 
     const coreProperties = {};
     coreProperties.serviceProviderProperties = this.serviceProviderProperties;
+
     this.setCore(new Core(coreProperties));
     this.core.addListener(listener);
     this.core.language = this.language;
@@ -542,7 +545,7 @@ export default class IntegrationModel {
     if (window.navigator.onLine) {
       this.core.editionProperties.dbclick = false;
       this.core.editionProperties.isNewElement = true;
-      this.core.openModalDialog(this.target, this.isIframe);
+      this.core.openModalDialog(this.target, this.isIframe, this.editorOptions);
     } else {
       let modal = document.getElementById("wrs_modal_offline");
       modal.style.display = "block";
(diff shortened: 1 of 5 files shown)


icaparros-at-wiris (approved):

packages/devkit/src/integrationmodel.js
```diff
@@ -12,7 +12,8 @@ import warnIcon from '../styles/icons/general/warn_icon.svg';  //eslint-disable-
  * @typedef {Object} IntegrationModelProperties
  * @property {string} configurationService - Configuration service path.
  * This parameter is needed to determine all services paths.
- * @property {HTMLElement} integrationModelProperties.target - HTML target.
+ * @property {HTMLElement} 
+ * .target - HTML target.
```
xripoll-at-wiris, line 16: I think this change was made by mistake.

xripoll-at-wiris (commented):
There is already a mechanism to pass configuration settings to the integration model when initializing the plugin. These are called `integrationModelProperties` and are passed along the properties of the host editor (e.g. with the TinyMCE integration, they are passed within the TInyMCE settings). Maybe you can put `forceHandMode` in there.

packages/devkit/src/integrationmodel.js
jgonzalez-at-wiris: Yes, you are right, I´ll fix it, thanks!!

PR #869, opened 2023-12-22
Fix formulas with '&' symbol not rendereting

carla-at-wiris:
## Description

Formulas containing the `&` and `"` symbols were breaking; causing console errors and breaking all the following formulas.
The cause was that the mentioned symbols were not being transformed into their HTML Entities, and it caused render issues.
This PR replaces the characters for their corresponding HTML Entity.

## Steps to reproduce

1. Create a broken equation for the viewer, such as an empty matrix with latex: 
    `$$\begin{pmatrix}&&&\\&&&\\&&&\\&&&\end{pmatrix}$$`
2. Insert equations after the latex.
3. Render the formulas.
4. No error console appears and all the formulas are properly rendered.

---

[#taskid 42230](https://wiris.kanbanize.com/ctrl_board/2/cards/42230/details/)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,10 @@ Last release of this project is was 20th of December 2023.
   - feat: Add method that allows the integration forcing the hand mode.
   - feat(viewer): decode safe mathml
 
+## Unreleased
+
+  - fix(viewer): Formulas containing '&' symbol are not rendered. #KB-42230
+
 ### 8.7.3 2023-11-22
 
   - fix(tinymce): non blur chemtype svg icon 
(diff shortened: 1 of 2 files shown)


xripoll-at-wiris:
How can I test this with the local changes to the viewer? By default all the demos use the wiris.net viewer, and the html-viewer demo I can't get it to render but also it doesn't produce any errors.

carla-at-wiris:
> How can I test this with the local changes to the viewer? By default all the demos use the wiris.net viewer, and the html-viewer demo I can't get it to render but also it doesn't produce any errors.

You should change the viewer that the demo is using. I believe it will be interesting to change the viewer we are using or have a command to specify which one we want to use (if possible) but I thought it was a longer development that should be done on its own task. 

As for reproducing the error, you should use the local viewer from the master branch. It also can be reproduced without having to use the local viewer. With the regular demos, you should click on the update button to render the formulas and then the console error appears and the formulas after the "broken" one should not render

xripoll-at-wiris (approved):

PR #886, opened 2024-02-03
fix(ck5): respect options passed to data controller on set

### LINKED ISSUE #885: [CKEditor 5] options passed to `setData` are ignored
## Description

CKEditor5 `setData` options are ignored

## Environment

`@wiris/mathtype-ckeditor5@8.7.2`

- CKEditor 5
- Any Framework
- Any Browser
- Any OS

## Steps to reproduce

*What are the steps needed to reproduce the bug?*

* Initialize a CKE5 editor with with the MathType plugin added
* Call `editor.setData("<p>new content</p>", { batchType: { isUndoable: true } })`
* Notice that the change cannot be undone

## Expected result

The change should be undoable

## Actual result

The change is not undoable

## Other details

This is because the monkey-patched `data.set` method is not passing the options on to the inner `set` method.

jonscheiding:
## Description

Make monkey-patched `data.set` and `data.get` respect options (and any other arguments that may be added some day)

Closes #885

### DIFF
--- a/packages/ckeditor5/src/plugin.js
+++ b/packages/ckeditor5/src/plugin.js
@@ -413,9 +413,9 @@ export default class MathType extends Plugin {
     /**
      * Hack to transform $$latex$$ into <math> in editor.getData()'s output.
      */
-    editor.data.get = (options) => {
-      let output = get.bind(editor.data)(options);
-      
+    editor.data.get = (...args) => {
+      let output = get.bind(editor.data)(...args);
+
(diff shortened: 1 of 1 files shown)


jguillen-at-wiris (approved):

jonscheiding:
@jguillen-at-wiris It actually occurred to me that I should switch this to use the event mechanism that CKE5 offers for stuff like this, that keeps it from interfering with other plugins that are also trying to use that mechanism.

carla-at-wiris:
This has been included on our latest release. Check the following Pull Request for more information #958

Thank you for your contribution!

PR #893, opened 2024-02-05
fix: use local viewer

icaparros-at-wiris:
## Description

Use local viewer

## Steps to reproduce

1. Open the viewer demo
2. Open the inspector
3. Check that the viewer import request is made on the next endpoint:
   http://localhost:4200/node_modules/@wiris/mathtype-viewer/dist/WIRISplugins.js?viewer=image

---

[#taskid 43631](https://wiris.kanbanize.com/ctrl_board/2/cards/43631/details/)


icaparros-at-wiris:
> When you change the viewer properties, how do you manage to re-render the formulas with those properties?

This is not working because we commented the Observer code. Currently, the you are not able to change the configuration of the viewer demo directly from the demo right panel. I  think that this is  out of the scope of this fix.

PR #900, opened 2024-02-15
fix: htmlSanitize stripping away valid MML tags

icc:
## Description

The DOMPurify dependency has been upgraded to the latest version. 
The previous version is old and lack support for certain tags etc. 
E.g. support for the mprescripts tag was added in  3.0.2

The newer version should be compatible with how DOMPurify is used today.

## Steps to reproduce

Edit this MathML in e.g. CKEditor5: 
```
<math xmlns="http://www.w3.org/1998/Math/MathML">
  <mmultiscripts>
    <msubsup>
      <mi>c</mi>
      <mi>h</mi>
      <mi>h</mi>
    </msubsup>
    <mprescripts/>
    <mi>h</mi>
    <mi>h</mi>
  </mmultiscripts>
</math>
```
and observe `mprescripts` tag is removed resulting in a different expression.

### DIFF
--- a/packages/devkit/package.json
+++ b/packages/devkit/package.json
@@ -48,7 +48,7 @@
     "webpack-cli": "^5.0.0"
   },
   "dependencies": {
-    "dompurify": "^2.3.9",
+    "dompurify": "^3.0.8",
     "raw-loader": "^4.0.2",
     "uuid": "^8.3.2"
   }


icc:
As a workaround until this is merged the parent project may override the dependency in `package.json` like this:
```
  "overrides": {
    "@wiris/mathtype-html-integration-devkit": {
      "dompurify": "^3.0.6"
    }
  }
```

carla-at-wiris:
This solution has been included in our latest release. Check the following Pull Requests to see the changes related to it: #941 and #926.

Thank you for noticing the issue and taking the time to write the solution!

PR #902, opened 2024-02-21
Fix error when double-click a plain text

xjiang-at-wiris:
## Description
In generic and froala, there is an error in the console: `Uncaught TypeError: Cannot read properties of null (reading 'split')` because when user do double click to a text, the system recognise it as formula image .

After that error is shown, the behavior of the modal is erratic and the user is blocked.

## Steps to reproduce

1. Enter to generic and froala staging enviroument.
2. Input a plain text.
3. Double click to that text.
4. Try to edit it with MT/CT. 
5. Ensure that there is no console error.
---

[#taskid 44249](https://wiris.kanbanize.com/ctrl_board/2/cards/44249/details/)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,7 @@ Last release of this project was 5th of February 2024.
 ### Unreleased
   - fix: Error when open a plain text with MT/CT editor. #KB-44250
   - feat: Add version text on html-viewer window property. #KB-43186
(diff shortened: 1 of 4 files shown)


jguillen-at-wiris (commented):
Why did we remove the following line from Generic but not from Froala?
super.doubleClickHandler(element);

I couldn't find an issue with the double-click event, but I don't understand the difference.

PR #911, opened 2024-03-05
Fix TypeError when insert a math equation into the first line

xjiang-at-wiris:
Prevent read Wirisformula class from node when don't have it

## Description

Fix TypeError when insert a math equation into the first line and edit it with MT/CT

## Steps to reproduce

- Insert a math equation into the first line
- Click on Insert math equation. Do not close the modal
- Click on Insert chemistry formula
- Check there is no console error
---

[#taskid 44268](https://wiris.kanbanize.com/ctrl_board/2/cards/44268/details/)


carla-at-wiris (changes requested):
The issue is still reproducible in staging.

carla-at-wiris (approved):

PR #925, opened 2024-03-26
Add DOMPurify exception

xjiang-at-wiris:
## Description

Add some DOMPurify exception in order to avoid error when inserting some formula.

## Steps to reproduce

- Open a CKEditor5 demo.
- Type the following equation: `<math xmlns="http://www.w3.org/1998/Math/MathML"><mstack charalign="center" stackalign="right"><mn>2</mn><msrow><mo>+</mo><mn>3</mn></msrow><msline> </msline><mn>5</mn></mstack></math>`
- Once inserted, copy the equation and paste it using the keyboard.
- Check the pasted formula is exactly the same as the original one

---

[#taskid 44372](https://wiris.kanbanize.com/ctrl_board/2/cards/44372/details/)

### DIFF
--- a/packages/devkit/src/util.js
+++ b/packages/devkit/src/util.js
@@ -403,8 +403,8 @@ export default class Util {
     let annotationRegex = /\<annotation.+\<\/annotation\>/
     // Get all the annotation content including the tags.
     let annotation = html.match(annotationRegex);
-    // Sanitize html code without removing the <semantics> and <annotation> tags.
-    html = DOMPurify.sanitize(html, { ADD_TAGS: ['semantics', 'annotation'], ADD_ATTR: ['linebreak', 'indentalign']});
+    // Sanitize html code without removing our supported MathML tags and attributes.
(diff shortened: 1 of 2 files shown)


packages/devkit/src/util.js
```diff
@@ -403,8 +403,8 @@ export default class Util {
     let annotationRegex = /\<annotation.+\<\/annotation\>/
     // Get all the annotation content including the tags.
     let annotation = html.match(annotationRegex);
-    // Sanitize html code without removing the <semantics> and <annotation> tags.
-    html = DOMPurify.sanitize(html, { ADD_TAGS: ['semantics', 'annotation'], ADD_ATTR: ['linebreak', 'indentalign']});
+    // Sanitize html code without removing some mathml tags and attributes.
```
carla-at-wiris, line 406: ### 
```suggestion
    // Sanitize html code without removing our supported MathML tags and attributes.
```

carla-at-wiris (changes requested):

carla-at-wiris (approved):

PR #961, opened 2024-05-09
fix: infite loop in some demo when deploy

xjiang-at-wiris:
## Description

With the current webpack configuration, when deploying demos for both TinyMCE 5 and TinyMCE 6, the entire editor gets copied over, including some unnecessary files. Due to caching mechanisms, webpack continuously detects changes in the source files of the TinyMCE editor, triggering continuous hot reloads.

This change will exclude those mentioned files during deployment, retaining only the minified files as they are the ones actually needed.

## Steps to reproduce

- Buid and run TinyMCE 5 and 6 demo muliple times
- Ensure there is no infinite loop
- Ensure both editor & Mathtype functionality works without error

## Changes.md
`fix: Exclude certain unnecessary files that are copied over during the deployment of TinyMCE 5 & 6 demos to prevent an infinite loop. #KB-44423`

----
[#taskid 44423](https://wiris.kanbanize.com/ctrl_board/2/cards/44423/details/)


carla-at-wiris (commented):
TinyMCE 5 & 6 work perfect!

The block still happens on CKEditor4, not on the 5 though.

xjiang-at-wiris:
> TinyMCE 5 & 6 work perfect!
> 
> The block still happens on CKEditor4, not on the 5 though.

thanks, fixed!

jgonzalez-at-wiris:
Would be interesting to complement your changes with a removal of the `dist`s folders so we make sure that directories are empty every time we start the project: https://github.com/johnagan/clean-webpack-plugin. CC @carla-at-wiris

usantos-at-wiris (approved):
Tested all HTML demos, and didn't find any issues! Do we know why these 3 and not the other editors ? It makes sense to change froala, generic, ck5, and the viewer to this new config ? Great job!

xjiang-at-wiris:
> Tested all HTML demos, and didn't find any issues! Do we know why these 3 and not the other editors ? It makes sense to change froala, generic, ck5, and the viewer to this new config ? Great job!

Good question!

The core error in the problem was that the dist folder was repeatedly written and mixed in when building and packaging the project, including some problematic files (For example, we need some code from Tiny to run Tiny editor in our demo).

So, the generic and viewer components didn't need any changes. They shouldn't have used any third-party assets from the beginning.

The architecture of CKEditor 5 and CKEditor 4 is completely different. So, for some unknown reason, we didn't need to include CK5 files in the dist folder to run the editor.

Froala is most likely to encounter a similar problem because it also packages the entire editor. However, fortunately, it didn't cause any errors. Of course, if the goal is to maintain cleanliness, Froala should be cleanable as well.

jgonzalez-at-wiris (approved):

carla-at-wiris (approved):

PR #984, opened 2024-06-21
Improve viewer handling errors

carla-at-wiris:
## Description

Improved the way we handle the errors and some other minor stuff. 

This is a side development due to the fact that there was appearing a 404 console error that we did not know how to handle.

## Steps to reproduce

Validate that the code makes sense.
Try out the viewer demo and verify that it works as expected.

---

[#taskid 48497](https://wiris.kanbanize.com/ctrl_board/2/cards/48497/details/)

### DIFF
--- a/packages/viewer/src/services.ts
+++ b/packages/viewer/src/services.ts
@@ -20,23 +20,19 @@ export class StatusError extends Error {
  * Helper function to process responses from the editor services.
  * These usually come wrapped in a JSON with a status field that can be either "ok" or "warning".
  * If status is "ok", return the result value along it. Otherwise, throw a StatusError.
- * @param {Promise<Response>} response - The response given by the service.
+ * @param {Response} response - The response given by the service.
  * @returns {Promise<any>} The unwrapped result of the response, if valid.
  * @throws {StatusError} Service responded with a non-ok status.
+ *                       If the response.json() fails, it'll return a generic exception.
  */
-export async function processJsonResponse(response: Promise<Response>): Promise<any> {
-  try {
-    const { status, result } = await (await response).json();
+async function processJsonResponse(response: Response): Promise<any> {
+  const { status, result } = await response.json();
 
-    if (status !== "ok") {
-      throw new StatusError("Service responded with a non-ok status");
-    }
-
-    return result;
-  } catch (e) {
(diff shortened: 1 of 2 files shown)


packages/viewer/src/services.ts
```diff
@@ -20,23 +20,19 @@ export class StatusError extends Error {
...
-      throw new StatusError("Service responded with a non-ok status");
-    }
+async function processJsonResponse(response: Response): Promise<any> {
+  const { status, result } = await response.json();
 
-    return result;
-  } catch (e) {
-    // TODO manage network and status non-ok errors
-    throw e;
+  if (status !== "ok") {
```
psala-at-wiris, line 31: This assumes that the caller will allways provide a valid response, but that is not the case. For example, in the function `createImage`: 

```typescript
  // this promise could be rejected
  const response = await callService(params, "showimage", MethodType.Post, url, extension);
  return processJsonResponse(response);
```

In this case, the rejected promise is handled by the parent caller `renderMathML` in a try/catch block, resulting in a `invalid MathML format` error message, which is confusing if the mathml is perfectly fine and the call fails due to a service error.

packages/viewer/src/services.ts
```diff
@@ -105,7 +96,7 @@ export async function mathml2accessible(mml: string, lang: string, url: string,
     ignoreStyles: "true",
   };
 
-  const response = callService(params, "service", MethodType.Post, url, extension);
+  const response = await callService(params, "service", MethodType.Post, url, extension);
```
psala-at-wiris, line 99: This is related to the previous comment.  
If we don't handle these promises, any resulting error messages from failed promises will come from the parent caller, resulting in somewhat confusing messages, which could make debugging very tricky.

usantos-at-wiris (approved):

xjiang-at-wiris (dismissed):

jgonzalez-at-wiris (dismissed):

packages/viewer/src/services.ts
psala-at-wiris: solved in the new version!

psala-at-wiris (approved):
Looks perfect after las changes!

jgonzalez-at-wiris (approved):

PR #985, opened 2024-06-21
fix(viewer): Fetch error on get configuration

carla-at-wiris:
## Description

During the QA process, a console error was spotted. 
The error is due to a URL fetching where, when it fails, we can assume that the client will be using cloud and when it does not fail, we can assume that they will be using on-premises with Java.

This error is not possible to be removed, since it's handled by the browser, but we still want to let the user know that it won't impact negatively in any way. That's why this PR adds a warning for the specified check.

## Steps to reproduce

1. Open the viewer demo and validate that the warning appears.

---


jgonzalez-at-wiris (approved):

usantos-at-wiris (approved):
Didn't test, but code looks good, and easy to understand.

PR #990, opened 2024-07-02
Kb 48270

xjiang-at-wiris:
## Description

Optimized the performance of the viewer by implementing the following:

1. Added an **Interaction Observer** to avoid rendering all functions on the page simultaneously. Instead, it renders only the currently visible functions and some that are just outside the viewport.
2. Added a **Mutation Observer** to render newly added functions individually instead of re-rendering the entire page. This also allows for the quick conversion of LaTeX to MathML.
3. Enabled **concurrent** rendering of multiple functions within a set upper limit, ensuring all visible functions on the page are rendered in parallel.

## Steps to reproduce

1. Launch viewer on local `yarn && nx build viewer && nx start html-viewer` or in [kitchen](https://integrations.wiris.kitchen/KB-48270/html/viewer/) 
2. Try to put a lot of formulas, you can use [this](https://drive.google.com/drive/u/2/folders/1QaoSRQX_BIXmR_kCaQXCIP7M5Av5LjiO)
3. Ensure all mencioned before are working
4. Try to modify new added parameters on table `simultaneousmml` `simultaneouslatex` `vieweroffset`

---

[#48270](https://wiris.kanbanize.com/ctrl_board/2/cards/48270/details/)


demos/html/viewer/src/app.js
```diff
@@ -30,10 +30,12 @@ document.getElementById("viewer_form").addEventListener("submit", () => {
...
     zoom: zoom.value,
+    vieweroffset:vieweroffset.value,
+    simultaneousmml:simultaneousmml.value,
+    simultaneouslatex:simultaneouslatex.value,
   };
 
   window.viewer.properties.config = config;
 
   document.getElementsByTagName("main")[0].innerHTML = document.getElementById("content").value;
-  window.com.wiris.js.JsPluginViewer.parseElement(document.getElementsByTagName("main")[0]);
```
xjiang-at-wiris, line 38: It has been removed because we no longer need "Render under request."

packages/viewer/src/mathml.ts
```diff
@@ -48,6 +48,41 @@ function decodeSafeMathML(root: HTMLElement) {
   }
 }
 
+export async function renderAMathML(properties: Properties, mathElement: MathMLElement): Promise<void> {
```
xjiang-at-wiris, line 51: I would like to delete the `renderMathML` function, but it seems to have a lot of implications and requires many document modifications. However, it appears worth discussing its feasibility.

xjiang-at-wiris (commented):
Some explanation

packages/viewer/docs/spec.md
```diff
@@ -112,6 +112,9 @@ The following table contains a specification of each of the properties.
...
 | zoom                       | The scale of the generated image.                                                                                                                                                                                                                                                                                                        | Frontend | Positive floating point number                  | 1                                       |
+| vieweroffset                       | Number of pixels to render in advance.                                                                                                                                                                                                                                                                                                        | Frontend | Positive floating point number                  | 200                                       |
```
carla-at-wiris, line 115: I'm missing maybe some information on what would that parameter be of use, because if I'm using the viewer for the first time, with that information I would not understand.

Maybe add something like "We render only the formulas on screen. This parameter defines the ..."

packages/viewer/docs/spec.md
```diff
@@ -112,6 +112,9 @@ The following table contains a specification of each of the properties.
...
+| simultaneousmml                       |  Mmax number of simultaneous MathML rendering petition can make at the same time.                                                                                                                                                                                                                                                                                                        | Frontend | Positive floating point number                  | 50                                       |
```
carla-at-wiris, line 116:
```suggestion
| simultaneousmml                       |  Max number of possible simultaneous MathML rendering petitions at the same time.                                                                                                                                                                                                                                                                                                        | Frontend | Positive floating point number                  | 50                                       |
```

packages/viewer/docs/spec.md
```diff
@@ -112,6 +112,9 @@ The following table contains a specification of each of the properties.
...
+| simultaneouslatex                       |  max number of simultaneous LaTeX rendering petition can make at the same time.                                                                                                                                                                                                                                                                                                        | Frontend | Positive floating point number                  | 50                                       |
```
carla-at-wiris, line 117:
```suggestion
| simultaneouslatex                       |  Max number of possible simultaneous LaTeX rendering petitions at the same time.                                                                                                                                                                                                                                                                                                        | Frontend | Positive floating point number                  | 50                                       |
```

packages/viewer/src/app.ts
```diff
@@ -41,49 +41,88 @@ async function main(w: Window): Promise<void> {
...
     if (w.viewer.isLoaded) return;
 
     w.viewer.isLoaded = true;
 
-    // First render
-    properties.render();
+    // Interaction Observer
+    const mathobserver = new IntersectionObserver(async (entries, observer) => {
+
+      let latexPromises = [];
```
carla-at-wiris, line 55: I see that you are using the `simultaneousmml` but I fail to find the place where the `simultaneouslatex` is being used. It seems then that the parameter is not providing any value to the viewer at this point. Is this intentional, or am I failing to see something?

packages/viewer/src/mathml.ts
carla-at-wiris: I would like so too. I would even consider this part of the development of the task, since the `renderMathML` is used for the retro-compatibility methods, which could benefit from the new performance improvements.

It would require a part of investigation and maybe adapt the current methods so that `renderMathML` can be properly replaced.

carla-at-wiris:
I'm getting the following console errors on the viewer demo:
image
image

packages/viewer/src/app.ts
psala-at-wiris: Same here! 
In Latex.ts, line 21: 
```typescript
 const maxLatexPetition = properties.simultaneousmml;
```
Maybe maxLatexPetition should be equal to properties.simultaneouslatex? 

And in Mathml.ts, maybe we should use the same mechanism in the `renderMathML` function? 
or I am missing something?

packages/viewer/src/app.ts
```diff
@@ -41,49 +41,88 @@ async function main(w: Window): Promise<void> {
...
+      }
+      await Promise.all(latexPromises);
+      await Promise.all(mmlPromises);
+    }, {
+      rootMargin: `${properties.vieweroffset}px`
+    });
+
+    const allElements = document.querySelectorAll('*');
+    window.addEventListener("load", function () {
+      let counter = 0;
```
psala-at-wiris, line 87: Is this counter used for something? 
Looks like a development tool XD

packages/viewer/src/app.ts
```diff
@@ -41,49 +41,88 @@ async function main(w: Window): Promise<void> {
...
+    const mathobserver = new IntersectionObserver(async (entries, observer) => {
+
+      let latexPromises = [];
+
+      const maxMmlPetition = properties.simultaneousmml;
+      let mmlPromises = [];
+      for (const entry of entries) {
+
+        if (entry.isIntersecting && entry.target instanceof MathMLElement && entry.target.matches('math')) {
+          const promise = renderAMathML(properties, entry.target);
```
psala-at-wiris, line 62: Maybe I am missing something, but: 
Why the renderAMathMl method does not implement the petition guard and renderLatex does? 
Could be possible to implement the guard logic in the same way for both methods? For me is kind of confusing that one does the guard mechanism and the other doesn't.

packages/viewer/src/app.ts
xjiang-at-wiris: My mistake! It should be `const maxLatexPetition = properties.simultaneouslatex;`
FIxed, thank you both!

packages/viewer/src/app.ts
xjiang-at-wiris: As we talked in private, renderAMathMl and renderLatex have some different behavior, that's why they don't share the same guard mechanism

usantos-at-wiris:
Hey, remember to add the decode so this work !  😊«math xmlns=¨http://www.w3.org/1998/Math/MathML¨»«mfrac»«mrow»«mi»r«/mi»«mi»e«/mi»«/mrow»«mi»w«/mi»«/mfrac»«mi»r«/mi»«/math»

PR #1083, opened 2025-02-24
fix: Add check for currentInstance existance

### LINKED ISSUE #829: [CKEditor5] MathType plugin throws error when `editor.sourceElement` is not defined
## Description

When using an editor that doesn't have the `sourceElement` property, the `MathType` plugin throws an error during the destroy process. 

We utilize various versions of editors, such as the headless editor in our internal features or the Multi-root editor, which do not include the `sourceElement` property. This results in error since the the`currentInstance` variable in the the`MathType` plugin is not defined (the plugin attempts to destroy `currentInstance`, which is `undefined`).

Reference: https://github.com/wiris/html-integrations/blob/stable/packages/ckeditor5/src/plugin.js#L68C5-L68C20 

It could be simply fixed by adding the `if` statement:
```
destroy() { // eslint-disable-line class-methods-use-this
    if( currentInstance ) {
        currentInstance.destroy();
    }
 }
 ```

## Environment

Latest

CKEditor5

## Steps to reproduce

1. Create [multi-root editor](https://ckeditor.com/docs/ckeditor5/latest/installation/getting-started/predefined-builds.html#multi-root-editor).
2. Add `MathType` to the editor plugins.
3. Call `editor.destroy()`.

## Expected result

The editor should be destroyed without errors.

## Actual result

The error is thrown.
`TypeError: Cannot read properties of undefined (reading 'removeEvents') at MathType.destroy`

### PR DESCRIPTION
**Pull request by carla-at-wiris**
## Description

In some cases, due to CKEditor's plugins like Multi-root, or random errors, `currentInstance` could be undefined. 

This causes, that methods which use `currentInstance` will fail if there's no error handling.

At this moment, the only method using that object is the `destroy()` one. This task should include an error handling for it.

- **Related Kanbanize Card:** [Link to KB-54684](https://wiris.kanbanize.com/ctrl_board/2/cards/54684/details/)
- **Related GitHub Issue:** Closes #829

## Type of Change

> Please delete options that are not relevant.

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] My code follows the style guidelines of this project ( Run `yarn lint` to check)
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (if applicable)
- [x] My changes generate no new warnings or errors
- [ ] I have added tests that prove my fix is effective or that my feature works (if applicable)
- [ ] New and existing unit tests pass locally with my changes

## How should be tested?

1. Manually set current instance to `undefined` or create a Multi-root CKEditor 5 demo with MathType.
2. Execute `editor.destroy()` on the console

No errors should appear and the editor should be destroyed

### CODE REVIEW

**Review by ncornaglia-at-wiris — DISMISSED** [review]

File: packages/ckeditor5/src/plugin.js [hunk 1]
```diff
@@ -61,8 +61,10 @@ export default class MathType extends Plugin {
    * Inherited from Plugin class: Executed when CKEditor5 is destroyed
    */
   destroy() {
-    // eslint-disable-line class-methods-use-this
-    currentInstance.destroy();
+    if (currentInstance) {
+      // eslint-disable-line class-methods-use-this
+      currentInstance.destroy();
+    }
```
jgonzalez-at-wiris, line 67: Would this work? Just to keep it a little bit cleaner.

```suggestion
      // eslint-disable-line class-methods-use-this
      currentInstance?.destroy();
```

carla-at-wiris: Yes for sure! I just committed your suggestion 😄

ncornaglia-at-wiris (approved):

jgonzalez-at-wiris (approved):

PR #1107, opened 2025-06-25
Change ckeditor5 language config propety check to handle null prototype

l-lejman:
Hi
Hi, I am a developer of CKEditor 5.

CKEditor5 uses a null-prototype object when the editor configuration is modified using the editor.config.define() function, in order to prevent prototype pollution vulnerabilities.

https://github.com/ckeditor/ckeditor5/blob/master/packages/ckeditor5-utils/src/config.ts#L198

MathType integration uses hasOwnProperty on the language object from the config, which causes an error when that property is defined using editor.config.define(), as the resulting object has a null prototype and does not include the hasOwnProperty method.

We need this change to prevent editor crashes when the language configuration is modified in this way.


carla-at-wiris:
Handled on https://github.com/wiris/html-integrations/pull/1116

